PR #2, opened 2018-03-01
AIDA-76

bdpham:
Made changes to config.simple_aida.json to display document table.

### DIFF
--- a/src/app/config/config.simple_aida.json
+++ b/src/app/config/config.simple_aida.json
@@ -139,7 +139,7 @@
 				"bindings": {
 					"title": "Wiki Viewer",
 					"database": "aida",
-					"table": "reified"
+					"table": "normal"
 				},
 				"row": 13,
 				"col": 0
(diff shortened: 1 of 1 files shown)


src/app/config/config.simple_aida.json
```diff
@@ -131,7 +139,7 @@
 				"bindings": {
 					"title": "Wiki Viewer",
 					"database": "aida",
-					"table": "test"
+					"table": "reified"
```
eaciv: Should the Wiki viewer be using the 'normal' table?

eaciv (commented):
Just the one comment. Let me know if it makes sense. If not, everything else looks good and I'm good to approve

src/app/config/config.simple_aida.json
bdpham: You're right. I forgot.

eaciv (approved):

PR #7, opened 2018-03-14
Aida 99 multiple filter fields and relations

bdpham:
Data-table component can have multiple filter fields. Added relations to simple-verdi config.

### DIFF
--- a/src/app/components/base-neon-component/base-neon.component.ts
+++ b/src/app/components/base-neon-component/base-neon.component.ts
@@ -588,6 +588,9 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
         });
     }
 
+    /**
+     * Returns FieldMetaData object with the given columnName
+     */
     getFieldObject(columnName: string, mappingKey?: string) : FieldMetaData {
         let me = this;
         let find = function(name) {
--- a/src/app/components/base-neon-component/base-neon.component.ts
+++ b/src/app/components/base-neon-component/base-neon.component.ts
@@ -591,7 +591,7 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
     /**
      * Returns FieldMetaData object with the given columnName
      */
-    getFieldObject(columnName: string, mappingKey?: string): FieldMetaData {
+    private getFieldObject(columnName: string, mappingKey?: string): FieldMetaData {
         let me = this;
         let find = function(name) {
             return _.find(me.meta.fields, function(field) {
--- a/src/app/components/base-neon-component/base-neon.component.ts
+++ b/src/app/components/base-neon-component/base-neon.component.ts
@@ -588,10 +588,7 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
         });
     }
 
-    /**
-     * Get field object from the key into the config options
-     */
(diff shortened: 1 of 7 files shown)


src/app/components/data-table/data-table.component.ts
```diff
@@ -609,8 +616,16 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
         this.executeQueryChain();
     }
 
-    removeFilter() {
-        this.filters = [];
+    removeFilter(filter: any) {
+        // We do it this way instead of using splice() because we have to replace filter array
```
eaciv: If splice is the right way to do this, you should still use it. Then you can create a copy of the array with either array.slice() or [...array].

src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -611,6 +611,27 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
         return fieldObject || this.datasetService.createBlankField();
     }
 
+    /**
+     * Get field object from the key into the config options
+     */
+    findFieldObjects(bindingKey: string, mappingKey?: string): FieldMetaData[] {
```
eaciv: Did you have to add this to base-neon or is this part of a Lorelei merge?

src/app/components/data-table/data-table.component.ts
```diff
@@ -654,26 +669,53 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
...
-            let key = dataField;
-            let prettyKey = this.active.dataField.prettyName;
-            let filter = {
-                id: undefined, // This will be set in the success callback of addNeonFilter.
-                key: key,
-                value: value,
-                prettyKey: prettyKey
-            };
-            this.addFilter(filter);
+            let object = this.active.rawData.filter((obj) =>
```
eaciv, line 668: Please add a couple comments to this section so it's easy to tell at a glance what logic we're using

src/app/config/config.simple_verdi.json
```diff
@@ -17,7 +17,21 @@
 					"name": "reified",
 					"prettyName": "Reified"
 				}]
-			}]
+			}],
+			"relations": [
+				{"members": [
```
eaciv, line 22: Did you ever get a good answer from Daniel about why "members" is required?

eaciv (changes requested):
Just a few questions and a request for comments

src/app/components/data-table/data-table.component.ts
bdpham: Ok. That makes sense. I pulled this piece from a different component.

src/app/components/base-neon-component/base-neon.component.ts
bdpham: I added it because I need an array of FieldMetaData to be returned. I'll look into possibly combining it with findFieldObject method.

src/app/components/data-table/data-table.component.ts
bdpham: Will do.

src/app/config/config.simple_verdi.json
bdpham: No. I never got around to asking him.

eaciv (approved):

PR #9, opened 2018-03-19
Aida 79

eaciv:
Use visjs for network graph


src/app/config/config.simple_verdi.json
```diff
@@ -1,14 +1,14 @@
...
+			"name": "VERDI Test",
 			"connectOnLoad": true,
-			"layout": "aida-layout",
+			"layout": "verdi-layout",
 			"datastore": "elasticsearch",
 			"hostname": "localhost",
 			"databases": [{
-				"name": "aida",
-				"prettyName": "AIDA",
+				"name": "verdi-rdf",
```
bdpham, line 10: With this change to the database name, does the createCurlFile from the AIDA repo reflect these changes and create the correct ElasticSearch database? If so, please provide the correct documentation in confluence (https://nextcentury.atlassian.net/wiki/spaces/AIDA/pages/198279178/Integration+with+Neon+Graph+Visualization) and provide the updated file in AIDA scripts.

eaciv: I updated the wiki and showed you how to create the index from the current script. I will update the script's default index in my AIDA-115.

bdpham (approved):

PR #12, opened 2018-03-28
Aida 94

gitnordt:


bdpham (approved):
Kafka offset monitor works. Refactoring is required on the media-viewer component, however that will be done in a different task (AIDA-78)

PR #15, opened 2018-04-11
Aida 78

gitnordt:


src/app/config/config.TA1_verdi.json
```diff
@@ -7,7 +7,7 @@
             "datastore": "elasticsearch",
             "hostname": "localhost",
             "databases": [{
-                "name": "verdi-rdf",
+                "name": "verdi",
```
bdpham: I'm not sure what we wanted to name the elasticsearch index. I was under the impression that we were going to use verdi-rdf for the reified data, and verdi for logs.

bdpham (approved):

PR #17, opened 2018-04-18
Aida 144

cedorman:
Made package.json use aida rather than lorelei;  added combined config.json (config.verdi.json)


bdpham:
How will the change to package.json affect the merge process with lorelei? Will we have to keep this commit off of our pushes to Lorelei-demo?

cedorman:
Yes, it will be incompatible with their package.json.  We don't have to have it, but using AIDA-Dash as-is on AWS doesn't work because of the hard-coded  lorelei paths that it has.


Clark



Clark Dorman, Chief Engineer
w (443) 545-3151 | clark.dorman@nextcentury.com
NEXTCENTURYCORPORATION | m 443.545.3100 | f 443.285.0799 | www.nextcentury.com<http://www.nextcentury.com/>

________________________________

bdpham (approved):

PR #19, opened 2018-05-24
Make it so you can click on documents one after another

cedorman:
Changed the config.verdi.json so that the filters in the document table do not apply to the document table (so they are still there) and you can only click one (multiple select turned off).  This means that you can go through the documents and select them one at a time without having to un-select the old one.

Example on EC2:   http://34.201.36.109/aida/


gitnordt (approved):
Jill mentioned that for Lorelei she only wanted to see one document when filter is on. I see that "ignoreSelf" allows you to switch this feature off and on. Great add!

PR #29, opened 2018-06-22
created script to automatically create Neon Dashboard "version"…

Charlotte-Hahn:
…for front end info tab. Updated link for "?" icon

### DIFF
--- a/scripts/set_version.sh
+++ b/scripts/set_version.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+# Get the current directory (in a safe / consistent way)
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+cd $DIR/..
+pwd
+
+#assign variable
+ngVersion=($(jq -r '.version' package.json | cut -c1-5))
+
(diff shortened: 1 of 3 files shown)


scripts/set_version.sh
```diff
@@ -0,0 +1,20 @@
...
+#assign variable
+ngVersion=($(jq -r '.version' package.json | cut -c1-5))
+
+gitId=$(git rev-parse HEAD | cut -c1-8)
+
+curDate=`date +%Y-%m-%d` 
+
+#concatenate variables
+FULL="${ngVersion}-SNAPSHOT-${gitId}-${curDate}"
+echo "{\"name\":\"neon-gtd\",\"version\":\""$FULL"\"}" > ./src/app/config/version.json
```
bdpham, line 18: Is there a reason why we are still calling it "neon-gtd"? In the demo, it says "Neon-Dashboard"

bdpham: Not sure how many changes there will be, or if it is worth the work, however, it might be a good idea to change any neon-gtd related names/variables to neon-dash to differentiate from the AngularJS version (neon-gtd).

bdpham (approved):

PR #34, opened 2018-07-02
Aida-238

bdpham:
Modified query-bar component to handle arrays in filters


gitnordt (approved):
Code review and tests passed

PR #42, opened 2018-07-11
Aida 244

gitnordt:


gitnordt:
This change can only be tested by changing the config.json file. Find the Thumbnail Grid in the config file and set the "showOnlyFiltered" parameter to false. This will allow all of the documents to be displayed including the one audio document.

Charlotte-Hahn (approved):

PR #45, opened 2018-07-19
Aida 243

gitnordt:


bdpham (changes requested):
app.component.spec.ts does not have the latest refactoring for ThumbnailDetails, therefore, tests fail.

bdpham (approved):

PR #48, opened 2018-07-26
Aida 278

Charlotte-Hahn:


gitnordt (approved):
Tests passed and changes work as expected

PR #50, opened 2018-08-01
Lorelei merge

gitnordt:


bdpham (commented):
Is there a reason why verdi-favicon was moved from assets/images to assets/spritesheets? verdi.config.json still seems to be referencing assets/images/verdi-favicon.ico but it's no longer there.

bdpham (approved):

PR #52, opened 2018-08-09
AIDA-311

eaciv:
Add options for node and link names.
Refactored adding nodes and edges to add nodes first and edges after.
Modified config.verdi.json to use new database.


gitnordt (approved):
Code review passed.

PR #53, opened 2018-08-10
Aida 207

gitnordt:


src/app/components/network-graph/network-graph.component.ts
```diff
@@ -616,14 +657,23 @@ export class NetworkGraphComponent extends BaseNeonComponent implements OnInit,
             for (let j = 0; j < nodes.length && graph.nodes.length < limit; j++) {
                 let nodeEntry = nodes[j];
                 if (this.isUniqueNode(nodeEntry)) {
-                    graph.addNode(new Node(nodeEntry, nodeNames[j], nodeName, 1, nodeColor, false, textColor));
+                    if (nodeColor.length > 1) {
+                        let index = categoryField.includes('EVENT') ? this.nodeCategories.indexOf('EVENT') :
```
bdpham: This is data and aida specific; is there any way we can take aida specific logic out?

bdpham:
The color mapping works, however, because of the aida specific logic, I don't think that it should be merged into master. One suggestion while we find a way to separate out the aida logic is to use this branch to create a branch specifically for the demo this weekend.

gitnordt:
@bdpham I added a config variable in order to perform the same functionality without the AIDA-specific category check.

bdpham (approved):

PR #55, opened 2018-08-22
Aida 334

jakevb:
On Verdi-Dash TA2, filter by the subject "voter turnout was over 70%".  The center node should be yellow and the other nodes should be blue.


eaciv:
Should be yellow, right?

jakevb:
You are right.  Just double checked your demo.  I was mistaken.

bdpham:
Can you pull master?

bdpham (approved):

PR #62, opened 2018-10-02
Added legend to network graph

gitnordt:
…lity to set the color scheme for nodes and edges, and legend label cleanup


ThomasSchellenbergNextCentury:
@snorpet We're going to want to call setColorListByLength (or ByIndex) in all of the visualizations that use the colorSchemeService now, right?  Else they'll default to using the first colorList (with 3 colors)?

ThomasSchellenbergNextCentury:
I fixed some lint errors.  Not sure how they got there, though, since the build is supposed to automatically fix the lint errors for you and fail if it can't... Anyway, it's fine now.

gitnordt:
@ThomasSchellenbergNextCentury > @snorpet We're going to want to call setColorListByLength (or ByIndex) in all of the visualizations that use the colorSchemeService now, right? Else they'll default to using the first colorList (with 3 colors)?

Yes, this is true. I changed it in the bar chart, but if you know of other places that are using the colorSchemeService, please feel free to make the change. It may be more useful to change it to an object and then use a key rather than an index to choose a color list. This effort was out of scope for my task, but possibly a good to add to the NEON ToDos.

ThomasSchellenbergNextCentury:
Sounds good.  I've created THOR-847 to address the issue (hopefully) next sprint.  Thanks!

ThomasSchellenbergNextCentury (approved):

bdpham (approved):

PR #66, opened 2018-10-12
Aida 378

bdpham:
added missing code from merge

### DIFF
--- a/src/app/components/network-graph/network-graph.component.ts
+++ b/src/app/components/network-graph/network-graph.component.ts
@@ -1101,7 +1101,7 @@ export class NetworkGraphComponent extends BaseNeonComponent implements OnInit, 
                         neon.query.where(filterField.field, '=', element));
                     singleFilter = this.createFilterObject(filterField.field, filterField.data.toString(), filterField.field);
                     clause = neon.query.or.apply(neon.query, clauses);
-                    filterArray.push({singleFilter, clause})
+                    filterArray.push({singleFilter, clause});
                 } else {
                     for (let data of filterField.data) {
                         singleFilter = this.createFilterObject(filterField.field, data, filterField.field, '=');
--- a/src/app/components/network-graph/network-graph.component.ts
+++ b/src/app/components/network-graph/network-graph.component.ts
@@ -489,20 +489,10 @@ export class NetworkGraphComponent extends BaseNeonComponent implements OnInit, 
         return null;
     }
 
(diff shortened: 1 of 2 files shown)


src/app/components/network-graph/network-graph.component.ts
```diff
@@ -1101,29 +1085,47 @@ export class NetworkGraphComponent extends BaseNeonComponent implements OnInit,
...
-            let field = selectedNode.isLink ? this.options.linkField : this.options.nodeField;
-            let myFilter = {
-                id: undefined,
-                field: field.columnName,
-                prettyField: field.prettyName,
-                value: nodeName,
-                operator: '='
-            };
-            this.addFilter(myFilter);
+            for (let filterField of selectedNode.filterFields) {
```
ThomasSchellenbergNextCentury, line 1112: Calling addNeonFilter within a loop (by calling addFilter on lines 1105 and 1111) will cause errors due to race conditions.  You need to call addNeonFilter once on your first filter, wait for a successful response from the backend, and then call addNeonFilter again on your second filter.  Right now, you're calling addNeonFilter in rapid-fire succession, which sends all the add-filter requests to the backend without waiting for the backend to process the requests and thus may get the backend filter state out-of-whack.

The addNeonFilter function has a 4th argument for a callback function.  You need to move the loop behavior to a separate function and then, on your first call to addNeonFilter supply a callback that calls your new function to add the next filter.  Let me know if you have questions or I'm not being clear.

ThomasSchellenbergNextCentury (changes requested):
You've got an issue with your filter-adding behavior.  Otherwise everything looks fine.

src/app/components/network-graph/network-graph.component.ts
bdpham: Ahhh that makes sense. I will fix that. Thanks.

gitnordt (changes requested):
I noticed that if you double-click on a node in the Network Graph to filter and then click on a hypothesis, the closeable filter in the network graph disappears as if the filter was removed. However if you look at the filter tray, the filter still exists. 
screenshot from 2018-10-12 13-56-07

bdpham:
@snorpet I will put in a different issue for the bug that you found.

ThomasSchellenbergNextCentury (approved):

gitnordt (commented):
That's acceptable. Let me know the bug number for the filtering issue.

PR #69, opened 2018-10-17
THOR-377 Warning for Deleting Save State

Konnoke:
Added Confirmation Dialog Component

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -240,6 +240,16 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
         });
     }
 
+    openCustomConnectionDialog() {
+        let config = new MatDialogConfig();
+        config.viewContainerRef = this.viewContainerRef;
+
+        this.customConnectionDialogRef = this.dialog.open(CustomConnectionComponent, config);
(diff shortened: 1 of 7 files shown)


src/app/app.component.ts
```diff
@@ -153,6 +154,37 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
 
     }
 
+    changeFavicon() {
```
ThomasSchellenbergNextCentury, line 157: Please confirm:  Most of these functions were simply reordered, and the only thing that changed was removing the openCustomConnectionDialog function, correct?

Konnoke: Thanks for pointing it out! OpenCustomConnectionDialog should not have been removed. I reordered the functions alphabetically but I forgot to added it back.

ThomasSchellenbergNextCentury: Thanks!  I can't tell you how much I appreciate ordering functions in alphabetical order, but you could also consider doing that ordering in its own commit in the future so that it's separated from the implementation of new behavior.  :)

ThomasSchellenbergNextCentury (approved):

PR #71, opened 2018-10-19
Lorelei merge 10-19-18

ThomasSchellenbergNextCentury:
Kevin, please don't merge this yourself.  I'd like to review it first but won't get time today.


Konnoke (approved):

src/app/components/map/map.component.ts
```diff
@@ -64,6 +64,7 @@ export class MapLayer extends BaseNeonLayer {
     public latitudeField: FieldMetaData;
     public longitudeField: FieldMetaData;
     public sizeField: FieldMetaData;
+    public hoverPopupField: FieldMetaData;
```
ThomasSchellenbergNextCentury, line 67: @sharknc In the future, whenever you add a new config option, it's be best to also add a mechanism to change that option within the UI.  In this case, you could add a mat-select in map.component.html for the hover popup field.  Right now, hoverPopupField can only be set in the config file.  (Note that, while we should do this each time we add new config options, we may still be missing some options here are there.)

src/app/components/media-viewer/media-viewer.component.ts
```diff
@@ -60,6 +60,7 @@ export class MediaViewerOptions extends BaseNeonOptions {
     public typeField: FieldMetaData;
     public typeMap: any;
     public url: string;
+    public autoplay: boolean;
```
ThomasSchellenbergNextCentury, line 63: @rartiss55 In the future, whenever you add a new config option, it's be best to also add a mechanism to change that option within the UI. In this case, you could add a mat-toggle-group in media-viewer.component.html for the autoplay option. Right now, autoplay can only be set in the config file. (Note that, while we should do this each time we add new config options, we may still be missing some options here are there.)

PR #72, opened 2018-10-22
Thor 864

Konnoke:
DO NOT MERGE. Just review


src/app/app.component.ts
```diff
@@ -286,11 +277,37 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
...
 
+    toggleAboutNeon() {
+        this.showRightPanel = true;
+        this.rightPanelService.changeActivePanel('aboutNeon', 'About Neon');
+        this.rightPanelService.showAboutNeon = true;
+    }
+
+    toggleAddVisualization() {
+        this.showRightPanel = true;
+        this.rightPanelService.changeActivePanel('add-vis', 'Visulization');
```
ThomasSchellenbergNextCentury: This is super nitpicky, but why is add-vis hyphenated while aboutNeon (line 282) is camel case?

src/app/app.component.html
```diff
@@ -50,25 +38,32 @@
...
-            <mat-icon class="neon-icon-large">settings</mat-icon>
-        </button>
-        <button mat-icon-button class="neon-icon-button-large" aria-label="About..." tooltip-position="below"
-            matTooltip="Show Application Information" (click)="showAbout = true; sidenavright.toggle()">
-            <mat-icon class="neon-icon-large">info</mat-icon>
+
+        <mat-menu #appMenu="matMenu" yPosition="below" [overlapTrigger]="false">
+            <button mat-menu-item (click)="toggleAddVisualization(); sidenavright.toggle()">
+                <mat-icon>add_circle</mat-icon>Visualizations</button>
+            <button mat-menu-item (click)="toggleDashboardLayouts(); sidenav.toggle()">
```
ThomasSchellenbergNextCentury: It would be clearer to rename the "sidenav" element as "sidenavleft" to better differentiate it from "sidenavright" (here, and on line 2).

src/app/app.component.ts
Konnoke: Good point. I'll change add-vis to camel case

src/app/app.component.ts
```diff
@@ -55,17 +56,17 @@ import { VisualizationContainerComponent } from './components/visualization-cont
 })
 export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
     @ViewChild(DashboardOptionsComponent) dashboardOptionsComponent: DashboardOptionsComponent;
+    @ViewChild(MatMenuTrigger) trigger: MatMenuTrigger;
     @ViewChild(NgGrid) grid: NgGrid;
     @ViewChildren(VisualizationContainerComponent) visualizations: QueryList<VisualizationContainerComponent>;
 
     @Input() sidenav = MatSidenav;
     // Used to determine which pane is show in the right sidenav
```
ThomasSchellenbergNextCentury, line 62: Is this actually used somewhere?  Am I missing it?

src/app/app.component.html
wagnerwj: sideNavleft - lets keep cases please.

src/app/app.component.ts
Konnoke: You're right, it's not being used right now. I'll remove it.

src/app/components/settings/settings.component.ts
```diff
@@ -0,0 +1,107 @@
...
+    public simpleSearch = {};
+    public simpleSearchField = {};
+
+    constructor(
+        public datasetService: DatasetService,
+        public exportService: ExportService,
+        public injector: Injector,
+        public rightPanelService: RightPanelService,
+        public themesService: ThemesService
+        ) {
```
ThomasSchellenbergNextCentury: The indentation here is weird.

src/app/components/settings/settings.component.ts
```diff
@@ -0,0 +1,107 @@
...
+import { RightPanelService } from '../../services/right-panel.service';
+import { ThemesService } from '../../services/themes.service';
+
+import { ConfirmationDialogComponent } from '../../components/confirmation-dialog/confirmation-dialog.component';
+
+import * as _ from 'lodash';
+import { BaseNeonOptions } from '../base-neon-component/base-neon.component';
+import { DatasetOptions } from '../../dataset';
+
+export class SettingsOptions extends BaseNeonOptions {
```
ThomasSchellenbergNextCentury: I don't think you need this SettingsOptions class...?

src/app/components/right-panel/right-panel.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    constructor(
+        public rightPanelService: RightPanelService,
+        public themesService: ThemesService
+    ) {
+        this.rightPanelService = rightPanelService;
+        this.themesService = themesService;
+    }
+
+    ngOnInit() {
+        //
```
ThomasSchellenbergNextCentury: I know that you're mid-progress on this task, so no rush, but I find it helpful to add a comment when you have an empty code block like this.  For example, I don't know whether this function still needs to be implemented, or is here to be overridden by subclasses, or just does nothing.  (Well, in this case it's an Angular function, so I have some idea, but you get the point!)

src/app/components/settings/settings.component.ts
```diff
@@ -0,0 +1,107 @@
...
+        newStateName: '',
+        stateToLoad: '',
+        stateToDelete: ''
+    };
+
+    public confirmDialogRef: MatDialogRef<ConfirmationDialogComponent>;
+    public exportTarget: string = 'all';
+    public options: SettingsOptions;
+    public simpleSearch = {};
+    public simpleSearchField = {};
```
ThomasSchellenbergNextCentury, line 59: As we discussed earlier, you can use the Neon Messenger to send custom events to the Simple Filter.  Here's a brief summary on using the Messenger:

- Use neon.eventing.Messenger() to create a messenger object.  Save it in the component.
- Use messenger.subscribe(channel, (message) => { /* your function */ }) to have your function called whenever a message is published over that channel.
- Use messenger.publish(channel, message) to publish a message over a channel.  The message is any object you want.

So you can create a custom event (something like "simple_filter_changed"), have the Simple Filter subscribe to those events, and have the Setting Component publish those events.  In the future, the Dataset Service can also publish those events in order to update the Simple Filter based on the config file.

src/app/components/settings/settings.component.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+@Component({
+  selector: 'app-settings',
+  templateUrl: 'settings.component.html',
+  styleUrls: ['settings.component.scss']
+})
+export class SettingsComponent implements OnInit {
+
+    @Input() sidenav: MatSidenav;
```
ThomasSchellenbergNextCentury: Will this be used in your upcoming changes, or should it be removed?

src/app/components/settings/settings.component.html
```diff
@@ -0,0 +1,52 @@
+<form class="settings-container" #optionsForm="ngForm">
+
+    <h3>
+        <span>Export Format</span>
+    </h3>
+
+    <p></p>
```
ThomasSchellenbergNextCentury: Two things:
1. I don't think you need the span tags.  If you do, just use the h3 tags and update their CSS as needed.
2. Remove the paragraph tags and add spacing in the CSS as needed (after the headers and before the dividers).  I think the paragraph tags are an anti-pattern that's currently used everywhere in our code when we should really just be using CSS instead.

Having fewer elements for each section (export, themes, display tools) means the fewer we have to duplicate when adding future sections.

src/app/components/right-panel/right-panel.component.html
```diff
@@ -0,0 +1,17 @@
+<div [class.neon-green-dark-theme]="themesService.getCurrentTheme().id === 'neon-green-dark-theme'">
+    <mat-toolbar class="title" color="primary">
+        <span>{{ rightPanelService.activePanel.title }}</span>
+    </mat-toolbar>
+
+    <div class="right-panel-container">
```
ThomasSchellenbergNextCentury: Does this need to be its own Component, or could it just be combined with the App Component?

I've been trying to decide whether this is useful as its own Component.  I think we'd only use it in the one place (the navbar).  It's also pretty tightly coupled with the navbar (if you add a button to the navbar in the App Component, you need to add a section for the panel in here too).  It would also make communication between the buttons and the panel easier (since it would all be done in the same Component).

What do you think?

src/app/services/right-panel.service.ts
```diff
@@ -0,0 +1,59 @@
...
+    //right panel
+    public showAboutNeon: boolean = false;
+    public showAddVis: boolean = false;
+    public showDashboardLayouts: boolean = false;
+    public showGear: boolean = false;
+    public showSaveState: boolean = false;
+    public showSettings: boolean = false;
+    //Toolbar
+    public showSimpleSearch: boolean = false;
+    public showVisShortcut: boolean = true;
```
ThomasSchellenbergNextCentury: As we discussed earlier, I think you can replace this Service with the Neon Messenger.

1. Move all the "show" booleans to the Component that shows the right panel (currently the Right Panel component, but maybe the App Component -- see my earlier comment).
2. Have this Component subscribe to a custom channel (maybe something like "show_right_panel") and update the appropriate "show" boolean.  The event message may be as simple as a string (about, add-vis, etc.)
3. Whatever currently changes the right panel (including the App Component and the individual Panel Components), rather than calling changeActivePanel or setting rightPanelService.show* to true, now publishes a message to the custom channel.
4. Since some of the Panel Components need the current status of some of the "show" booleans, they should subscribe to the custom channel too and update their own internal "show" variables.

Cons:
- We have to duplicate a few of the "show" variables across a few of the Components.
- Become dependent on Neon's Messenger (though I don't think it is going to change anytime soon).

Pros:
- Reduce Service bloat.
- Move all of the panel-toggling behavior into the same Component.
- I think it will be easier to extend the panel-toggling behavior (if needed in the future) within the subscription function than with just toggling booleans on a service.
- Don't have to think of a better name than "RightPanelService" :)

What do you think?

src/app/components/right-panel/right-panel.component.html
Konnoke: Initially I had it in the app component but I decided to move it to its own component because it made the app component html very cluttered and thought it would make implementing the gear option object easier in the future. But yea separating it created the problem of communication between the panel and the app component which I created the service to address..

src/app/services/right-panel.service.ts
```diff
@@ -0,0 +1,59 @@
...
+import * as neon from 'neon-framework';
+
+class ActivePanel {
+    public type: string;
+    public title: string;
+}
+
+@Injectable()
+export class RightPanelService {
+    private activePanel: ActivePanel;
```
ThomasSchellenbergNextCentury: Do we really need the activePanel?  It looks like most of the panel-toggling behavior is handled by the "show" booleans below.  Right now the activePanel is only storing the panel title which I think could be done without a custom class.  Am I missing something?

src/app/components/right-panel/right-panel.component.html
ThomasSchellenbergNextCentury: I think that the visualization options ("gear") would be its own panel like About, Add Visualization, or Settings.  Would that make sense based on your own understanding of the changes needed to move the visualization options into the Right Panel?

src/app/components/settings/settings.component.html
```diff
@@ -0,0 +1,52 @@
...
+                    <mat-select name="simple-search-field" placeholder="Simple Search Field" [(ngModel)]="simpleSearchField" required="true"
+                        [disabled]="options.fields.length == 0">
+                        <mat-option [value]="emptyField">(None)</mat-option>
+                        <mat-option *ngFor="let field of options.fields" [value]="field">{{ field.prettyName }}</mat-option>
+                    </mat-select>
+                </mat-form-field>
+            </div>
+        </div>
+        <div class="checkmark-container">
+            <input type="checkbox" [checked]="rightPanelService.showVisShortcut" (change)="rightPanelService.showVisShortcut = !rightPanelService.showVisShortcut">
```
ThomasSchellenbergNextCentury: This is Sophie's fault, not your's, but I find it a bit frustrating how many places you can toggle showVisShortcut...

src/app/components/add-visualization-panel/add-visualization-panel.component.ts
```diff
@@ -0,0 +1,68 @@
...
+import { neonVisualizations } from '../../neon-namespaces';
+import { RightPanelService } from '../../services/right-panel.service';
+import { ThemesService } from '../../services/themes.service';
+
+@Component({
+    selector: 'app-add-visualization-panel',
+    templateUrl: 'add-visualization-panel.component.html',
+    styleUrls: ['add-visualization-panel.component.scss']
+})
+export class AddVisualizationPanelComponent implements OnInit {
```
ThomasSchellenbergNextCentury: This is somewhat minor, but I'm bothered by the naming inconsistency here:  We have AboutNeon and Settings, and then we have AddVisualizationPanel (with the Panel at the end).  I know that we already have an AddVisualizationComponent, but I'd ask that we either replace it with this newest version OR add "Panel" to the end of the other Component names.

ThomasSchellenbergNextCentury:
Hey @Konnoke I know I made a lot of comments and requested some changes but I wanted to say that overall this looks REALLY good.  You've done a great job getting a lot done in a short amount of time and I think that you've made some insightful design choices.  Kudos to you!

src/app/components/add-visualization-panel/add-visualization-panel.component.ts
Konnoke: The idea was to eventually replace Add Visualization with Add Visualization Panel because I didn't want to delete Add Visualization in case we wanted to keep it. I did get confirmation from Sophie yesterday that we won't need the old add visualization component so I'll go ahead and replace it.

src/app/services/right-panel.service.ts
Konnoke: Yea that's leftover from a different design that I initially used where all of the addVis, aboutNeon, setting, and eventually gear shared a superclass. I scrapped that idea because addvis, aboutneon, and setting were different enough to warrant their own component so I'll keep that superclass idea for just the gear option.

src/app/app.component.html
ThomasSchellenbergNextCentury: In that case, we should probably change the right one to sideNavRight too :)

src/app/components/right-panel/right-panel.component.html
Konnoke: There were 2 designs that I was thinking with the visualization options (gear):

1. The first one was one where the right panel component would take in the gear object (baseNeonOptions)/or gear component and would create its own template from the gear options which originally warranted the need for the right panel component. 

2. The second one is the one you mentioned where the gear would be its own panel, which eliminates the need for the right panel component. 

The first one was intertwined with the common superclass idea I originally tried to implement for the addVis, aboutNeon, settings, and gear but scrapped.

ThomasSchellenbergNextCentury: To me, it seems like we're implementing the same behavior for both 1 and 2, and the question is where to implement that behavior:  in the Right Panel (1) or in a sub-Panel (2).  I'd argue for 2.

src/app/services/right-panel.service.ts
Konnoke: Just to confirm, for 2-4, I would NOT need to add any code to neon (server/middleware) right? All new codes will just be on the front end?

ThomasSchellenbergNextCentury: That's correct.  Here's an example of publishing and subscribing to custom events (in this case, select_id):  https://github.com/NextCenturyCorporation/neon-dash-internal/blob/master/src/app/components/base-neon-component/base-neon.component.ts#L939

src/app/app.component.ts
```diff
@@ -286,11 +300,38 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
...
+        this.resetPanel();
+        this.rightPanelTitle = 'Visulization';
+        this.showAddVis = true;
+
+    }
+
+    toggleDashboardLayouts() {
+        //this.resetPanel();
+        //this.rightPanelTitle = 'Dashboard Layouts';
+        //this.showDashboardLayouts = true;
```
ThomasSchellenbergNextCentury: Remind me, what's the status on the dashboard layouts panel?

Konnoke: That's blocked by THOR-826 and the new config rework

src/app/services/dataset.service.ts
```diff
@@ -176,6 +176,30 @@ export class DatasetService {
...
+        this.dataset.options.simpleFilter.fieldName = simpleField.columnName;
+    }
+
+    /**
+     * returns the simple search field
+     * @return {Object}
+     */
+    public getActiveDatasetSimpleFilterFieldName() {
+        return this.dataset.options.simpleFilter.fieldName;
+    }
```
ThomasSchellenbergNextCentury, line 208: Is there a reason why we're saving the simple filter field name in the dataset service?  It looks like it's only referenced in the settings component... Am I missing something?

Konnoke: If simple field is in the config then it will exist in the dataset service. Previously, the only place simpleField was set was from the config at the start. The setting component allows the simple filter to be changed from the original simple filter that was set from the config. 

One thing I noticed with this code is this assume that a simple filter already exists, I might add a check to create the simplefilter if it didn't exist in the config.

ThomasSchellenbergNextCentury:
@Konnoke , at some point you should write unit tests for the new elements and features you added.  I'm not sure if you'd prefer to do this now or closer to the end of this Epic.  I've compiled a list of areas to test:

App
- Check that it's showing the correct elements (including ngIfs) by default
- Check that toggling the show* booleans updates the correct shown elements
- Check that calling the reset* and toggle* functions updates the correct booleans and title
- Check that the messenger subscribes to the correct channels and that the callbacks update the correct booleans

Add Visualization
- Check that the messenger subscribes to the correct channels and that the callbacks update the correct booleans
- Check that the publish function updates the correct booleans and publishes the correct data to the correct channels

Settings
- Check everything lol

Simple Filter
- Check that the messenger subscribes to the correct channels and that the callbacks update the correct booleans
- Check that the publish function updates the correct booleans and publishes the correct data to the correct channels

For the messenger, I'm hoping that you can either mock it (using Angular tools or just creating a new class that extends the Messenger) or access its underlying channels.  We can discuss different strategies sometime.

src/app/services/dataset.service.ts
ThomasSchellenbergNextCentury: Ah, config file compatibility -- right.  I forgot.  Thanks!

And, yes, check for a null object and create a new object if needed.

src/app/app.component.ts
ThomasSchellenbergNextCentury: Right, OK.  Can you please add a TODO comment containing the ticket number so we know why this isn't working yet / is commented out?

src/app/services/dataset.service.ts
```diff
@@ -176,6 +176,30 @@ export class DatasetService {
...
+     */
+    public setActiveDatasetSimpleFilterFieldName(simpleField: FieldMetaData) {
+        this.dataset.options.simpleFilter.fieldName = simpleField.columnName;
+    }
+
+    /**
+     * returns the simple search field
+     * @return {Object}
+     */
+    public getActiveDatasetSimpleFilterFieldName() {
```
ThomasSchellenbergNextCentury: 1. Please change line 189 "@return {Object}" to "@return {FieldMetaData}"
2. Please add a return type to line 191 like "public getActiveDatasetSimpleFilterFieldName(): FieldMetaData {"

ThomasSchellenbergNextCentury:
I realized that I was currently the only reviewer on this PR (well, besides Billy, but I think he's just supervising), so I added Rachel too for a second set of eyes.  She'll review everything after lunch.

src/app/app.component.html
```diff
@@ -1,47 +1,54 @@
...
+            </mat-toolbar>
+        
+            <div class="right-panel-container">
+                <div class="about-neon-container" *ngIf="showAboutNeon">
+                    <app-about-neon></app-about-neon>
+                </div>
+                <div class="add-visualization-container" *ngIf="showAddVis">
+                    <app-add-visualization></app-add-visualization>
+                </div>
+                <div class="saved-state" *ngIf="showSaveState">
```
ThomasSchellenbergNextCentury: This is SUPER nitpicky but "saved-state" is the only class without "-container" at the end...

src/app/services/dataset.service.ts
Konnoke: I realized it should be string instead because fieldName in SimpleFilter is just a string and not fieldMetaData.

Konnoke:
One thing I was debating is if I should combine all the show booleans into a single object or string variable that has the type (aboutNeon, addVis, etc) and the ngIf refers to a function that checks the object for its respective type. 

Pro:

-  Combines of all of the show booleans into one string/object
- Gets rid of the resetPanel function.

Con: 

- There will be a function for each respective panel type to check the string/object

ThomasSchellenbergNextCentury:
Re:  combine all the show booleans into a single string

I think that's a good idea.  A string named "currentPanel" or something.

Your function can just be an "equals" function:

```javascript
equals(one: any, two: any): boolean {
    return one === two;
}
```

```html
*ngIf="equals(currentPanel, 'addVis')"
```

src/app/app.component.ts
```diff
@@ -54,22 +55,30 @@ import { VisualizationContainerComponent } from './components/visualization-cont
     ]
 })
 export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
-    @ViewChild(DashboardOptionsComponent) dashboardOptionsComponent: DashboardOptionsComponent;
     @ViewChild(NgGrid) grid: NgGrid;
+    @ViewChild(SaveStateComponent) SaveStateComponent: SaveStateComponent;
```
rartiss55: nitpick - property name should start with a lowercase letter.

src/app/app.component.ts
```diff
@@ -286,11 +300,38 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
...
-            this.dashboardOptionsComponent.loadStateNames();
+    toggleAboutNeon() {
+        this.resetPanel();
+        this.rightPanelTitle = 'About Neon';
+        this.showAboutNeon = true;
+    }
+
+    toggleAddVisualization() {
+        this.resetPanel();
+        this.rightPanelTitle = 'Visulization';
```
rartiss55: Typo -- should say "Visu**a**lization"

src/app/components/settings/settings.component.html
```diff
@@ -0,0 +1,40 @@
+<form class="settings-container" #optionsForm="ngForm">
+    <h3 class="header"> Export Format</h3>
+    <app-export-control [exportTarget]="exportTarget"></app-export-control>
+    
+    <mat-divider></mat-divider>
```
ThomasSchellenbergNextCentury: Can you please add a margin around the mat-divider elements too?  Something like "margin:  15px 0;" looked good (IMO).

src/app/app.component.html
```diff
@@ -1,47 +1,54 @@
...
+            <mat-toolbar class="title" color="primary">
+                <span>{{ rightPanelTitle }}</span>
+            </mat-toolbar>
+        
+            <div class="right-panel-container">
+                <div class="about-neon-container" *ngIf="showAboutNeon">
+                    <app-about-neon></app-about-neon>
+                </div>
+                <div class="add-visualization-container" *ngIf="showAddVis">
+                    <app-add-visualization></app-add-visualization>
```
ThomasSchellenbergNextCentury, line 21: What happened to that alternative view of the Add Visualization sidenav that had all the icons but smaller and in a single column without labels?  Do you remember that from the mockup?  Should that be a separate ticket in this Epic?

src/app/app.component.scss
```diff
@@ -71,3 +71,14 @@
...
+
+.right-panel-container {
+    height: 100%;
+    width: 410px;
+}
+
+.title{
+   position: -webkit-sticky; /* Safari */
+   position: sticky;
+   top: 0;
```
rartiss55, line 83: I'm seeing a bug where if you have the Add Visualizations panel open, the items within it overlap over the title as you scroll. I think you could set the z-index value here to fix that, but perhaps there's a better way.

src/app/components/settings/settings.component.ts
```diff
@@ -0,0 +1,158 @@
...
+        stateToDelete: ''
+    };
+
+    public confirmDialogRef: MatDialogRef<ConfirmationDialogComponent>;
+    public exportTarget: string = 'all';
+    public options;
+    public searchField: FieldMetaData;
+    public showVisShortcut: boolean = true;
+    public showSimpleSearch: boolean;
+    public simpleFilter = new BehaviorSubject<SimpleFilter>(undefined);
```
rartiss55, line 57: Is this simpleFilter BehaviorSubject property still needed since we're using the neon.eventing.Messenger()?

src/app/app.component.html
Konnoke: Ah that's the toolbar which will be a future ticket. Sophie told me not to worry about that for this task because there's still some stuff that needs to be figured out for that.

src/app/app.component.scss
Konnoke: I encounter that bug too but it's not consistent. if I refresh the page and scroll then sometimes it doesn't overlap.

src/app/components/simple-filter/simple-filter.component.ts
```diff
@@ -13,7 +13,7 @@
  * limitations under the License.
```
rartiss55, line 13: I'm getting a test failure for SimpleFilter tests: 'should not show in the UI when the configuration does not include a simpleFilter option'. 

(Sidenote: I also see a test failure for data-table tests, but I think that will be fixed with merging).

ThomasSchellenbergNextCentury: Yeah, I'm also seeing HeadlessChrome disconnected during the tests, which indicates that another error is happening.

src/app/app.component.scss
Konnoke: I set a zindex of 9999 for the title and I don't see the bug anymore. I haven't seen any other problems with it so far.

src/app/components/settings/settings.component.ts
Konnoke: It is needed for now because the check and the simpleFilter component uses the BehaviorSubject.

Over the weekend, I attempted to replace the behavior subject simpleFilter in the simple filter component to use the neon event messenger. That involved adding a publish function in the dataset service which was called in the dataset selector app and/or functions within the dataset service that would indicate when the dashboard is selected or first loading but broke creating filters in the simpleFilter component. FilterId in the simpleFilter component is also a behavior subject. Getting rid of using behavior subject in the simpleFilter component will require rewriting how filters are created or removed in the simpleFilter component.

Why does the behavior subject matter for the simpleFilter component one might ask? Well there is the issue of consistency but one use case is broken. If you try to create a simpleFilter on a dashboard where there is no simpleFilter set from the config, it will not work. That's because the simpleFilter in the setting component and dataset service uses messenger, but in the simpleFilter component uses behavior subject. When a simpleFilter object is created in the dataset, the simpleFilter behavior subject in the simpleFilter component does not know the new value. One solution is to use behavior subject in the dataset service to update the behavior subject but that means that we are using both behavior subject and the neon event messenger in the dataset service. A better solution might be to get rid of using behavior subject all together in the simpleFilter component which I attempted over the weekend but that requires more effort than the one I put into over the weekend. 

Tl:dr, Merge the PR for now and I'll work on this on a separate branch after the merge.

rartiss55: Thanks for investigating this! Yeah I think we can worry about this later, or perhaps it will be resolved with some of the Neon 3 rework.

src/app/components/simple-filter/simple-filter.component.ts
rartiss55: It looks like this test got commented out -- could we either look into fixing it now or making a separate ticket to fix if the issue is more involved?

rartiss55: So I think for this, the issue is there is now a simpleFilter even if it doesn't exist in the config? So maybe here, in order to have element be null, we'd need to set the boolean showSimpleSearch to false:

```javascript
constructor(mockDataset = true, showSimpleSearch = true) {
         TestBed.configureTestingModule({
             declarations: [
                 SimpleFilterComponent
...
         let fixture = TestBed.createComponent(SimpleFilterComponent);
         this.fixture = fixture;
         this.component = fixture.componentInstance;
         this.component.showSimpleSearch = showSimpleSearch;
         this.filterService = this.getInjected(FilterService);
         this.detectChanges();
...

describe('Component: SimpleFilter unconfigured', () => {
    let tester: SimpleFilterTester;
beforeEach(() => tester = new SimpleFilterTester(false, false));

it('**should not show in the UI when showSimpleFilter is set to false**', () => {
    expect(tester.element).toBeFalsy();
});
```

Thoughts?

rartiss55:
@ThomasSchellenbergNextCentury I updated the tickets @Konnoke made to cover your unit test comments:
https://nextcentury.atlassian.net/browse/THOR-905
https://nextcentury.atlassian.net/browse/THOR-906

rartiss55 (approved):

ThomasSchellenbergNextCentury (approved):

PR #75, opened 2018-10-23
Moved getExportFields from the BaseNeonComponent to the BaseNeonOptions

ThomasSchellenbergNextCentury:
PLEASE DO NOT MERGE YET.  Just approve the changes and let me merge when ready.

THOR-870

Moved getExportFields in the Components to the Options objects.  Trying to make the Component less complicated by compartmentalizing config-related behavior in the Options and moving common behavior to the Superclass Options.  Now the Superclass Options will automatically return all non-empty Fields with getExportFields and any Subclass Options that want custom behavior can override that method.

Please let me know if you have any questions.

Thanks!


Konnoke:
Sounds like a good idea

Konnoke (approved):

PR #76, opened 2018-10-23
Replaced custom functions in Map with deepFind

ThomasSchellenbergNextCentury:
I replaced our custom functions in the Map Component with the deepFind function in neonUtilities that we use throughout the rest of the code.


ThomasSchellenbergNextCentury:
Also @sharknc if you could please ensure that these changes also work with the data you are currently using, I would appreciate it.

sharknc:
@ThomasSchellenbergNextCentury will do

sharknc (approved):

PR #79, opened 2018-10-25
THOR-847 THOR-854 ColorSchemeService color list simplification

ThomasSchellenbergNextCentury:
First in a series of changes to the ColorSchemeService.

In this change, I've removed the overly complicated behavior of managing multiple color lists.  I've updated the single color list to use the new color theme designed by UX.  Right now we're just using the Teal theme colors but eventually the colors will change based on the selected theme.

Stacey, I've also removed the custom color behavior in the Network Graph.  Please make sure that this change still works with the AIDA data or if we need to add new custom behavior.  I think eventually we should allow you to set custom node colors using the config file but that can be a future task.

### DIFF
--- a/src/app/components/aggregation/aggregation.component.spec.ts
+++ b/src/app/components/aggregation/aggregation.component.spec.ts
@@ -53,9 +53,8 @@ describe('Component: Aggregation', () => {
     let fixture: ComponentFixture<AggregationComponent>;
     let getService = (type: any) => fixture.debugElement.injector.get(type);
 
-    let COLOR_1 = new Color(173, 216, 230);
-    let COLOR_2 = new Color(228, 26, 28);
-    let COLOR_3 = new Color(55, 126, 184);
+    let COLOR_1 = new Color(255, 135, 55);
+    let COLOR_2 = new Color(94, 80, 143);
 
     initializeTestBed({
         declarations: [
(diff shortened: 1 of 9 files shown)


gitnordt (changes requested):
AIDA will still need a way to apply custom colors to the network graph. The colors listed in the ColorSet are unappealing and make it difficult to read the node text. 

unappealingcolors

Do we need to modify the config file and network graph to include access to these colors?
        [   //3 colors - defined palette	
            new Color(173, 216, 230),	
            new Color(223, 159, 159),	
            new Color(255, 218, 153)	
        ]

ThomasSchellenbergNextCentury:
@snorpet I think the ideal solution would be to use the config file to set custom color lists that the network graph (or other specific widgets) would use instead of the default theme-based colors.

Would you like that implemented before we merge-in this change?  (It may not be too hard -- I can take a look tomorrow.)

gitnordt:
@thomasSchellenbergNextCentury I believe I had the colors set in the config before I realized we were using a color service. I might be able to find the code fairly easily.

ThomasSchellenbergNextCentury:
@snorpet I think there's a colorMaps option in the config but I don't think anything in the code is actually using it right now...

gitnordt:
> @ThomasSchellenbergNextCentury I believe I had the colors set in the config before I realized we were using a color service. I might be able to find the code fairly easily.



> @snorpet I think there's a colorMaps option in the config but I don't think anything in the code is actually using it right now...

@ThomasSchellenbergNextCentury  I will put in a separate ticket for this one and make the changes.

gitnordt (approved):

PR #81, opened 2018-10-29
Aida 363

eaciv:
Add object metadata for network graph in lieu of storing node metadata and edge metadata separately.

### DIFF
--- a/src/app/components/network-graph/network-graph.component.html
+++ b/src/app/components/network-graph/network-graph.component.html
@@ -128,41 +128,71 @@
                 <p></p>
 
                 <!-- Color Options -->
-                <mat-form-field *ngIf="!options.isReified">
-                    <input matInput type="color" placeholder="Node Color" [(ngModel)]="options.nodeColor" [value]="options.nodeColor" (ngModelChange)="reloadGraph()">
-                  </mat-form-field>
-                <p></p>
+                <div *ngIf="!options.isReified">
+                    <mat-form-field>
+                        <input matInput type="color" placeholder="Node Color" [(ngModel)]="options.nodeColor"
+                               [value]="options.nodeColor" (ngModelChange)="reloadGraph()">
+                    </mat-form-field>
+                    <p></p>
 
-                <mat-form-field *ngIf="!options.isReified">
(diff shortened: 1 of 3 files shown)


src/app/components/network-graph/network-graph.component.html
```diff
@@ -78,6 +78,15 @@
                     </mat-form-field>
                     <p></p>
 
+                    <mat-form-field>
+                        <mat-select placeholder="Target Name Field" [(ngModel)]="options.targetNameField" required="false" (ngModelChange)="handleChangeData()"
+                                    [disabled]="options.fields.length == 0">
+                            <mat-option [value]="emptyField">(None)</mat-option>
+                            <mat-option *ngFor="let field of options.fields" [value]="field">{{ field.prettyName }}</mat-option>
+                        </mat-select>
+                    </mat-form-field>
```
ThomasSchellenbergNextCentury, line 87: Should we add a mat-select for the targetColorField too?

src/app/components/network-graph/network-graph.component.html
```diff
@@ -104,17 +113,8 @@
...
                     <p></p>
                 </div>
-
-                <mat-form-field>
-                    <mat-select placeholder="Node Shape" [(ngModel)]="options.nodeShape" required="false" (ngModelChange)="handleChangeData()"
-                                [disabled]="nodeShapes.length == 0">
-                        <mat-option [value]="emptyField">(Default)</mat-option>
-                        <mat-option *ngFor="let shape of nodeShapes" [value]="shape">{{ shape }}</mat-option>
-                    </mat-select>
-                </mat-form-field>
```
ThomasSchellenbergNextCentury, line 114: It looks like this was removed because it was duplicated (I'm assuming due to a bad merge) -- is this correct?

src/app/components/network-graph/network-graph.component.html
eaciv: Yep. I thought about adding this, but didn't see any color options. Looks like they're further down :/. I'll add it and refactor the color section.

src/app/components/network-graph/network-graph.component.html
eaciv: It was removed because it was duplicated. I can't speak to why it was in there in the first place.

ThomasSchellenbergNextCentury (approved):

PR #84, opened 2018-10-30
Feature/thor 859

sharknc:
- Updated map component to persist all unique id's associated with a MapPoint drawn on the map. This includes a single point or a collapsed point which contains multiple points at a single lat/long
- Updated map component to store a Map object of all hoverPopup values for all points for a given MapPoint that is drawn on the map.
- Created functionality to collapse duplicate hoverPopupValues that are redundant and display a count next to each. For example, if there are 10 points at the same lat/long with a hover popup value of "Medicine", the value displayed when hovering over that point on the map will be "Medicine(10)"
- Fixed all tests that were broken when implementing above functionality


src/app/components/map/map.component.ts
```diff
@@ -639,27 +655,34 @@ export class MapComponent extends BaseLayeredNeonComponent implements OnInit, On
...
+
         let map = new Map<string, UniqueLocationPoint>();
 
         for (let point of data) {
             let lngCoord = this.convertToFloatIfString(neonUtilities.deepFind(point, lngField)),
                 latCoord = this.convertToFloatIfString(neonUtilities.deepFind(point, latField)),
-                hoverPopupValue = neonUtilities.deepFind(point, hoverPopupField),
-                colorValue = colorField && point[colorField];
+                colorValue = colorField && point[colorField],
+                idValue = idField && point[idField];  //value must be top level nost nested array
```
ThomasSchellenbergNextCentury: Why can't the ID be a nested value?  I can understand if you need it to be a SINGLE value (and NOT an array), but I don't see why it shouldn't also be able to accept NESTED values.

What about something like this?

```javascript
let idValue = idField ? neonUtilities.deepFind(point, idField) : '';
idValue = idValue instanceof Array ? (idValue.length ? idValue[0] : '') : idValue;
```

ThomasSchellenbergNextCentury: As a side note, I don't see any reason why colorValue isn't using deepFind either.

src/app/components/map/map.component.ts
```diff
@@ -1020,6 +1062,35 @@ export class MapComponent extends BaseLayeredNeonComponent implements OnInit, On
...
+            let tableExists = false;
+            this.options.layers.forEach((elem) => {
+                if (eventMessage.database === elem.database.name) { dbExists = true; }
+                if (eventMessage.table === elem.table.name) { tableExists = true; }
+            });
+
+            if (dbExists && tableExists) {
+                this.options.id = Array.isArray(eventMessage.id) ? eventMessage.id[0] : eventMessage.id;
+                if (this.options.id !== this.previousId) {
+                    //this.tabsAndMedia = [];
```
ThomasSchellenbergNextCentury: Please remove this comment.

src/app/components/map/map.component.ts
```diff
@@ -1020,6 +1062,35 @@ export class MapComponent extends BaseLayeredNeonComponent implements OnInit, On
...
+                    this.previousId = this.options.id;
+                    this.executeQueryChain(0);
```
ThomasSchellenbergNextCentury: The argument for executeQueryChain is the layerIndex, which isn't always 0.  Track the layer on which the database and table exist using your loop above.

src/app/components/map/map.type.leaflet.ts
```diff
@@ -85,26 +85,68 @@ export class LeafletNeonMap extends AbstractMap {
 
     addPoints(points: MapPoint[], layer?: MapLayer, cluster?: boolean) {
         let group = this.getGroup(layer);
+
         for (let point of points) {
-            let circlOptions = {
+
+            let circlOptions = {};
```
ThomasSchellenbergNextCentury: Apparently whoever was here before you forgot the E in circle...

src/app/components/map/map.component.ts
sharknc: I believe I just followed the same pattern as other components for the idField and the colorValue was existing code. I will update to use deepFind.

src/app/components/map/map.type.leaflet.ts
```diff
@@ -85,26 +85,68 @@ export class LeafletNeonMap extends AbstractMap {
...
+                }
+            } else {
+                //default circle options
+                circlOptions = {
                     color: point.cssColorString === whiteString ? 'gray' : point.cssColorString,
                     fillColor: point.cssColorString,
                     weight: 1,
                     colorByField: point.colorByField,
                     colorByValue: point.colorByValue,
                     radius: Math.min(Math.floor(6 * Math.pow(point.count, .5)), 30) // Default is 10
```
ThomasSchellenbergNextCentury: Couldn't we combine the three definitions of circleOptions?  Something like this?

```javascript
let mapIsSelected = this.mapOptions.id && point.idValue;
let pointIsSelected = point.idList.includes(this.mapOptions.id);
circleOptions = {
                        color: point.cssColorString === whiteString ? 'gray' : point.cssColorString,
                        colorByField: point.colorByField,
                        colorByValue: point.colorByValue,
                        fill: mapIsSelected && pointIsSelected ? true : undefined,
                        fillColor: point.cssColorString,
                        fillOpacity: mapIsSelected ? (pointIsSelected ? 1 : .1) : undefined,
                        opacity: mapIsSelected && !pointIsSelected ? .2 : undefined,
                        radius: Math.min(Math.floor(6 * Math.pow(point.count, .5)), 30), // Default is 10
                        stroke: mapIsSelected && pointIsSelected ? false : undefined,
                        weight: mapIsSelected && pointIsSelected ? undefined : 1
};
```

sharknc: Sure thing

src/app/components/map/map.type.leaflet.ts
sharknc: Yea...or maybe the e was just one letter too long. I can update it

wagnerwj: Circl doesn't have an E in it!

src/app/components/map/map.component.spec.ts
```diff
@@ -47,6 +47,7 @@ import * as neon from 'neon-framework';
 import { DatasetServiceMock } from '../../../testUtils/MockServices/DatasetServiceMock';
 import { FilterServiceMock } from '../../../testUtils/MockServices/FilterServiceMock';
 import { initializeTestBed } from '../../../testUtils/initializeTestBed';
+import { stringify } from '@angular/core/src/util';
```
ThomasSchellenbergNextCentury: I don't think this needs to be imported.

src/app/components/map/map.type.leaflet.ts
ThomasSchellenbergNextCentury: Or do you think the redundancy is preferable to having inline conditionals here?

src/app/components/map/map.component.ts
ThomasSchellenbergNextCentury: Huh, interesting.  In this case, I think using deepFind is correct.

src/app/components/map/map.component.spec.ts
sharknc: I'm not sure I imported that. I was pulling changes in from lorelei-master, was could this be from a different commit?

src/app/components/map/map.type.leaflet.ts
ThomasSchellenbergNextCentury: Whoops, didn't see your reply.  Nevermind then.

sharknc: I didn't attempt t refactor because I was confused on what was going on and thought keeping it explicit may make it easier for the next person to track. After looking at your code though it does look much cleaner.

sharknc: This code does not seem to work. I couldn't figure out what is going on in the few minutes I spent looking at it. I'll have to look at the leaflet spec tomorrow and verify that its okay with setting fields to undefined.

src/app/components/map/map.component.ts
sharknc: updated the code, but it made me question why we are tracking the previous id in the first place. Is this to account for duplicate messages being sent? I'll commit my basic implementation so you can check it out.

src/app/components/map/map.component.ts
sharknc: fixed

src/app/components/map/map.component.ts
sharknc: fixed

src/app/components/map/map.component.ts
sharknc: I should note, I referenced the media viewer component when implementing this method, maybe that was particular to that component.

src/app/components/map/map.component.ts
sharknc: Updated both idValue and colorValue to handle edge case of an array being returned from deepFind(). Pushed changes

PR #91, opened 2018-11-06
Color Cleanup

ThomasSchellenbergNextCentury:
1. Moved the Color and ColorSet classes from the ColorSchemeService into their own file.
2. Changed the ColorSchemeService to use a database/table/field key rather than just a field name.
3. Updated the LegendComponent to use colorKeys rather than fieldNames.
4. Updated the components that use the ColorSchemeService and Legend as needed.


Konnoke (approved):
Opposite array

PR #92, opened 2018-11-06
Thor 885

chrislong:
Added X and Y axis labels to aggregate charts.

### DIFF
--- a/src/app/components/aggregation/aggregation.component.ts
+++ b/src/app/components/aggregation/aggregation.component.ts
@@ -827,30 +827,6 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
     }
 
     /**
-     * Returns whether the given subcomponent type is scaled.
-     *
-     * @arg {string} type
-     * @return {boolean}
-     */
-    isScaled(type: string): boolean {
-        switch (type) {
-            case 'bar-h':
-            case 'bar-v':
-            case 'histogram':
-            case 'line':
-            case 'line-xy':
-            case 'scatter':
-            case 'scatter-xy':
-                return true;
-            case 'doughnut':
-            case 'list':
-            case 'pie':
-            default:
-                return false;
-        }
-    }
-
-    /**
      * Returns whether the given subcomponent type has axes (e.g., line does,
      * pie does not).
      *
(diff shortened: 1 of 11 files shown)


src/app/components/aggregation/aggregation.component.ts
```diff
@@ -820,6 +850,31 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
...
     }
 
+    /**
+     * Returns whether the given subcomponent type has axes (e.g., line does,
+     * pie does not).
+     *
+     * @arg {string} type
+     * @return {boolean}
+     */
+    hasAxes(type: string): boolean {
```
ThomasSchellenbergNextCentury, line 836: This is the same as the isScaled function, so you can use that function (or replace it with this function, if you think hasAxes is a more descriptive name).

chrislong: Ok. I saw isScaled but I didn't know what it was for, so I wasn't sure it would always be the same (e.g., if more chart types were added later). I'll consolidate them.

ThomasSchellenbergNextCentury (changes requested):
I see a few errors in the unit tests:
```
Component: Aggregation options.createBindings does set expected properties in bindings FAILED
Component: Aggregation does show elements in sidenav options menu that have expected options FAILED
Component: Aggregation with config does show elements in sidenav options menu that have expected options from config FAILED
Component: Aggregation with XY config does show elements in sidenav options menu that have expected options if subcomponent type is XY FAILED
Component: Aggregation with date config does show elements in sidenav options menu that have expected options if X field is date type FAILED
```
The first one broke because you added the new axisLabelX and axisLabelY bindings, so you should be able to fix it by adding them to the test.

The others broke because you added the new input elements to the options menu.  You should be able to fix them by updating the checks for the input elements.

I know the test output is a bit strange (the tests are run twice -- once in Firefox and once in Headless Chrome -- and I find the Headless Chrome error stacktraces to be more helpful) so please let me know if you have any questions.  Thanks!

ThomasSchellenbergNextCentury (approved):

ThomasSchellenbergNextCentury:
Whoops, I didn't realize that this was merging into master.  I merged it into lorelei-master instead.

PR #93, opened 2018-11-07
Thor 865

Konnoke:
- Moved the dashboard layout from the left sidenav to the right sidenav
- Set the default value of currentPanel to DashboardLayouts so the dataset selector component is created and loads the dashboard.
- Added a catch for an error in the app component for the filter builder.


rartiss55:
Looks like this needs an additional UX improvement (having the dashboards layout popup appear as open on page load if no dashboard is selected as the default in the config), so holding off on approving until that's implemented.

Konnoke:
If no dashboard is selected as the defaults in the config, the dashboard layout popup and stays open now

rartiss55 (approved):

ThomasSchellenbergNextCentury (approved):

PR #98, opened 2018-11-13
Aida 416

gitnordt:


src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -499,6 +500,35 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
     }
 
+    addCountsToTaxonomy(data: any, groups: any[]) {
+        for (let group of groups) {
+            let count = 0;
+            group.nodeIds = [];
+            data.forEach((d) => {
+                let id = neonUtilities.deepFind(d, this.options.idField.columnName);
+                if (neonUtilities.deepFind(d, group.description).includes(group.name) && !group.nodeIds.includes(id)) {
+                    group.nodeIds.push(neonUtilities.deepFind(d, this.options.idField.columnName));
```
ThomasSchellenbergNextCentury: Can't you just push `id` here?

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -499,6 +500,35 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
+
+            if (group.hasOwnProperty('children')) {
+                this.addCountsToTaxonomy(data, group.children);
+                let childCount = 0;
+                for (let child of group.children) {
+                    childCount += child.nodeCount;
+                }
+
+                if (!group.nodeCount) {
+                    group.nodeCount = childCount;
```
ThomasSchellenbergNextCentury, line 525: This seems to be counting only the direct children and not other descendants (grandchildren, great grandchildren, etc.).  I wanted to ensure that this was intentional.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
gitnordt: Great catch! I made the id variable after the fact and must have forgotten to paste it here.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
gitnordt: Line 518 calls the function recursively so that descendants are counted as well.

ThomasSchellenbergNextCentury: Oh, whoops, I misunderstood then.

But isn't the original count (set on line 515) getting overridden here?

bdpham:
@snorpet The total number in the taxonomy viewer should reflect the total nodes in the network graph correct?

gitnordt:
> @snorpet The total number in the taxonomy viewer should reflect the total nodes in the network graph correct?

We discussed this issue in stand-up. Due to the fact that our data has multiple categories and types, some of the counts will be a little off. The network graph counts a node only once while the taxonomy counts the same node for multiple categories and types. In the taxonomy, there is no way to easily identify that a node is already listed under another category. As long as it's off by 2 or 3, Clark said that this is acceptable. It should be fixed whenever there is newly structured data available that has a single category and a single type.

ThomasSchellenbergNextCentury (approved):

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
gitnordt: No it's not overriding the counts and when checking the counts against the ES index they are accurate. I believe it is Javascript's object memory allocation that allows for this. If you would like for this to be done another way, I am open to suggestions?

bdpham (approved):

PR #104, opened 2018-11-19
Modify TA1 config for demo. Fix filter service bug

eaciv:
Replacing db, table, and field all at once was causing problems if one of the items was contained in the other. For instance, if the db was called "solid" and you wanted to replace all fields labeled "id", the "id" part of "solid" was replaced. Attempted to make this more explicit and controlled.

### DIFF
--- a/src/app/services/filter.service.ts
+++ b/src/app/services/filter.service.ts
@@ -47,7 +47,7 @@ export class FilterService {
     constructor(
         protected errorNotificationService: ErrorNotificationService,
         protected datasetService: DatasetService
-    ) {}
+    ) { }
 
     protected getDatabaseFilterState(onSuccess: (filterList: any[]) => any, onError: (response: any) => any) {
         neon.query.Filter.getFilterState('*', '*', onSuccess, onError);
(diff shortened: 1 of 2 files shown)


src/app/services/filter.service.ts
```diff
@@ -409,7 +409,9 @@ export class FilterService {
             } else {
                 Object.keys(object).forEach((key) => {
                     if (typeof object[key] === 'string') {
-                        object[key] = object[key].replace(oldDb, newDb).replace(oldTable, newTable).replace(oldField, newField);
+                        //TODO: why replacing all fields with all new values?!
+                        let val = object[key];
+                        object[key] = val === oldDb ? newDb : val === oldTable ? newTable : val === oldField ? newField : val;
```
ThomasSchellenbergNextCentury, line 414: You're right, all of this behavior is hacky.  I've created a ticket to review it when we make our upcoming changes to filter management:  https://nextcentury.atlassian.net/browse/THOR-939

No action required on your part.

ThomasSchellenbergNextCentury (approved):

PR #107, opened 2018-11-26
THOR-909 Replaced BaseNeonOptions with WidgetOption / WidgetOptionCollection

ThomasSchellenbergNextCentury:
Goals:
- Change the widget options from being flat primitive properties on BaseNeonOptions to being a collection of WidgetOption objects that wrap the primitive properties along with their corresponding information (label, dropdown options, etc.)
- Be able to send the WidgetOptions to the upcoming shared widget-options-menu component so it can dynamically create the dropdowns/inputs/toggles/etc.
- Remove the existing BaseNeonOptions classes and the excessive HTML for the unique options in the individual widgets
- Be able to extend the WidgetOptionsCollection with options from specific visualization types (like different thumbnail-grid views or aggregation charts) or projects (LORELEI- or AIDA-specific options)

Changes:
1. Created WidgetOptionCollection, WidgetOption, and option subclasses (see widget-option.ts)
2. Removed BaseNeonOptions, BaseNeonLayer, and BaseNeonMultiLayerOptions from BaseNeonComponent / BaseLayeredNeonComponent.  Moved many of their functions to BaseNeonComponent / BaseLayeredNeonComponent.
3. Added a WidgetOptionCollection (called "options") to BaseNeonComponent / BaseLayeredNeonComponent.
4. Created new abstract functions in BaseNeonComponent / BaseLayeredNeonComponent:  createFieldOptions, createLayerFieldOptions, createLayerNonFieldOptions, createNonFieldOptions, getWidgetDefaultLimit, getWidgetName
5. Updated all widgets to adopt the new changes.  Added some backwards compatibility support for old bindings.
6. Deleted the deprecated bar-chart, line-chart, scatter-plot, and timeline because I didn't want to spend 2+ hours updating them.

Notes:
- I (mostly) haven't touched the existing widget (gear) options menus in the HTML because that is part of the creation of the upcoming shared widget-options-menu component.


Konnoke:
Can we keep the old timeline? It was better and faster at handing bigger data than the aggregation one. Or is it too much of a monster to maintain and it's better to redirect the effort towards improving the aggregation histogram.

ThomasSchellenbergNextCentury:
@Konnoke So you're saying that the performance of the D3 timeline was better than that of the ChartJS histogram?

We can keep the old one for now, but it will eventually need to be replaced by an Aggregation Component using a D3 Histogram Subcomponent (which right now is probably blocked by THOR-904).

Konnoke:
@ThomasSchellenbergNextCentury Ah I didn't know there was a plan for a D3 Histogram Subcomponent. In that case, we don't have to keep the old timeline.

ThomasSchellenbergNextCentury:
@Konnoke Well, there is a PLAN, but that doesn't mean it's actually going to HAPPEN for another few months... :)

src/app/widget-option.ts
```diff
@@ -0,0 +1,381 @@
...
+    FIELD = 'FIELD',
+    FIELD_ARRAY = 'FIELD_ARRAY',
+    FREE_TEXT = 'FREE_TEXT',
+    MULTIPLE_SELECT = 'MULTIPLE_SELECT',
+    NON_PRIMITIVE = 'NON_PRIMITIVE',
+    SELECT = 'SELECT',
+    TABLE = 'TABLE'
+}
+
+export abstract class WidgetOption {
```
ThomasSchellenbergNextCentury, line 34: I wrote the WidgetOption as an abstract class with a set of subclasses (rather than a single class) in order to emphasize the types of options I expected.  The subclasses are more of a convenience to reduce the need to enter unneeded constructor arguments.

src/app/components/data-table/data-table.component.ts
```diff
@@ -247,6 +141,106 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
...
+    }
+
+    /**
+     * Returns the name for the unique widget.
+     *
+     * @return {string}
+     * @override
+     */
+    getWidgetName(): string {
+        return 'Text Cloud';
```
rartiss55, line 241: Was this intentional or was this meant for the text cloud component?

src/app/widget-option.ts
```diff
@@ -0,0 +1,381 @@
...
+
+    /**
+     * @constructor
+     * @arg {OptionType} optionType
+     * @arg {boolean} isRequired
+     * @arg {string} bindingKey
+     * @arg {string} prettyName
+     * @arg {any} valueDefault
+     * @arg {OptionChoice[]} valueChoices
+     * @arg {boolean|OptionCallback} [enableInMenu=true]
```
ThomasSchellenbergNextCentury, line 45: This will be a discussion for UX:  If an option is not currently needed/used (because of other options), should it be hidden, or just disabled?

src/app/components/data-table/data-table.component.ts
ThomasSchellenbergNextCentury: This was a copy-paste error.  Good catch!

ThomasSchellenbergNextCentury:
@rartiss55 I fixed the bug with the map not loading properly, but since I fixed it AFTER merging lorelei-master into this branch locally, I don't want to push my commits on this branch or else the PR is going to get messed up.  So I've pushed the fix to a new branch and I guess you can cherry pick it to test it locally?  Sorry for the inconvenience!  https://github.com/NextCenturyCorporation/neon-dash-internal/commit/ccbe5e8f0408de2be2bab54fc0c652975a0c3bc1

rartiss55:
@ThomasSchellenbergNextCentury - Tested with a diff of that commit mentioned and it looks like its working fine now. Thanks!

rartiss55 (approved):

Konnoke (approved):

PR #110, opened 2018-11-28
Thor 905

Konnoke:
Added Unit Tests for the App, Add Visualization, and simple filter component,

**Other Changes:**

- Fixed the filter builder icon path for ec2 (hopefully)
- Added an option to show or hide the filter builder icon in the settings panel.
- Added binding methods for the messengers to use 
- FilterBuilderIcon is dependent on the correct boolean in the app component html, before it was using the one for the simple search.


.gitignore
```diff
@@ -65,4 +65,3 @@ src/app/shared/index\.ts
 
 # Automatically-generated Derby database for saved states and stuff.
 savedStates/*
-*~
```
chrislong, line 68: Do you have a reason for removing this? It's for emacs backup files.

src/app/components/add-visualization/add-visualization.component.spec.ts
```diff
@@ -67,30 +74,68 @@ class TestAddVisualizationComponent extends AddVisualizationComponent {
     // TODO Add any needed custom functions here.
 }
 
-describe('Component: Sample', () => {
+describe('Component: Add Visualization', () => {
     let component: TestAddVisualizationComponent;
-    let fixture: ComponentFixture<TestAddVisualizationComponent>;
-    let getService = (type: any) => fixture.debugElement.injector.get(type);
+    let fixture: ComponentFixture<TestAddVisualizationComponent>,
+        getService = (type: any) => fixture.debugElement.injector.get(type);
```
chrislong: <nitpick>Why separate `let` for some variables but not others? I don't care which way, it just seems inconsistent.</nitpick>

src/app/components/aggregation/aggregation.component.html
```diff
@@ -228,35 +228,21 @@
...
-                               (keyup.enter)="refreshVisualization()">
-                    </mat-form-field>
-                    <p></p>
-
-                    <mat-form-field>
-                        <input matInput placeholder="Y axis Label" [(ngModel)]="options.axisLabelY" required="false"
-                               (keyup.enter)="refreshVisualization()">
-                    </mat-form-field>
-                    <p></p>
-                </div>
```
chrislong: Is there a reason the axis label options were removed?

src/app/components/add-visualization/add-visualization.component.spec.ts
ThomasSchellenbergNextCentury: It's currently by personal preference.  Looks like there's a one-variable-per-declaration setting that we could add to the typescript lint config.  I'm not sure if there's a best practice, but I'd think that having a single variable per declaration would be easier to maintain.

.gitignore
ThomasSchellenbergNextCentury: Obviously you should be using vim instead!  :)

ThomasSchellenbergNextCentury:
OK I think I fixed all the weird merge issues.  Sorry if some of the spacing is messed up.

@Konnoke Please checkout the changes and ensure everything still works as expected.

@chrislong Feel free to take a second look now.

e2e/typings.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference path="../typings/index.d.ts" />
```
ThomasSchellenbergNextCentury, line 1: Kevin, was this something you did, or is this another weird lingering merge thing?

src/app/app.component.spec.ts
```diff
@@ -166,14 +171,76 @@ describe('App', () => {
...
+
+    }));
+
+    it('toggle filter builder', async(() => {
+        component.showFilterBuilderIcon = false;
+        expect(debugElement.nativeElement.querySelectorAll('app-filter-builder').length === 0).toBeTruthy();
+        component.showFilterBuilderIcon = true;
+        component.createFilterBuilder = true;
+        component.openFilterBuilderDialog();
+        expect(debugElement.nativeElement.querySelectorAll('app-filter-builder')).toBeTruthy();
```
ThomasSchellenbergNextCentury, line 212: Do you need to check `.length` here?  I think querySelectorAll always returns an array and would thus always be truthy, right?

e2e/typings.d.ts
Konnoke: Weird lingering merge thing

src/app/app.component.ts
```diff
@@ -417,12 +433,9 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
     }
 
     ngOnInit(): void {
-        this.messenger.subscribe('showSimpleSearch', (message) => {
-            this.showSimpleSearch = message.showSimpleSearch;
-        });
-        this.messenger.subscribe('showVisShortcut', (message) => {
-            this.showVisShortcut = message.showVisShortcut;
-        });
+        this.messenger.subscribe('showSimpleSearch', (message) => this.bindShowSimpleSearch(message));
```
ThomasSchellenbergNextCentury: @Konnoke So this is OK and it obviously works but I think it would be better to do it this way:
```javascript
this.messenger.subscribe('showSimpleSearch', this.bindShowSimpleSearch.bind(this));
```
It's a little less code ~~and you're also not creating an extra function this way~~.  Does that make sense?

Edit:  Whoops; `bind` also creates a new function, so that's not a distinct benefit.  However, I do prefer this way (passing the function using `bind(this)`) because it is a bit safer since, in order to update the behavior in the future, you have to go to the `bindShowSimpleSearch` function and update it there.  On the contrary, with your current design, you can update the behavior either in `bindShowSimpleSearch` OR the anonymous function `(message) => this.bindShowSimpleSearch(message)` which introduces unneeded complexity.

src/app/app.component.ts
```diff
@@ -231,6 +235,18 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
         this.widgetGridItems.push(widgetGridItem);
     }
 
+    bindShowSimpleSearch(message) {
```
ThomasSchellenbergNextCentury: I'd recommend renaming these functions without the "bind" part (so, in this case, "showSimpleSearch").

src/app/app.component.ts
ThomasSchellenbergNextCentury: BTW Here's the documentation on bind:  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

ThomasSchellenbergNextCentury: @Konnoke Huh, I wonder if this will actually work based on how you're testing the subscriptions:  https://github.com/NextCenturyCorporation/neon-dash-internal/pull/110/files#diff-d5d984b316bea5c337b5e6f7b6448f62R224

Ugh, you may just have to keep the current behavior.  Why do you have to make everything so complicated, Kevin?!  (jk)

chrislong:
Looks ok to me.

src/app/components/add-visualization/add-visualization.component.spec.ts
```diff
@@ -81,13 +93,46 @@ describe('Component: AddVisualization', () => {
...
         component = fixture.componentInstance;
+        spyOnInit = spyOn(component, 'ngOnInit');
         fixture.detectChanges();
+
+        debugElement = fixture.debugElement;
     });
 
+    it('exists', (() => {
+        expect(component).toBeTruthy();
+    }));
```
ThomasSchellenbergNextCentury, line 111: You don't need to test that the component exists since all of your other unit tests will already do that for you!  I think these unit tests were left over in suites for other components but they're only really useful for stubs; when you add new (real) unit tests you can take these out.

src/app/app.component.spec.ts
```diff
@@ -166,14 +171,76 @@ describe('App', () => {
...
+
+        component.ngOnInit();
+        expect(spyOnMessengerSubscribe.calls.count()).toEqual(3);
+        component.bindShowSimpleSearch(message);
+        component.bindShowVisShortcut(message);
+        component.bindShowFilterBuilderIcon(message);
+
+        expect(spyOnBindShowFilterBuilderIcon.calls.count()).toEqual(1);
+        expect(spyOnBingShowSimpleSearch.calls.count()).toEqual(1);
+        expect(spyOnBingShowVisualShortcut.calls.count()).toEqual(1);
```
ThomasSchellenbergNextCentury: @Konnoke FYI You can also use the spy to validate that the arguments passed to a function are correct:
```javascript
  expect(spyOnBingShowSimpleSearch.calls.argsFor(0)).toEqual([{
    showSimpleSearch: true
  }]);
```
The argument for the `argsFor` function is the zero-based index of the call (so the arguments for the first time the function is called are returned with `argsFor(0)`).  It returns an array containing each function argument.

src/app/app.component.spec.ts
```diff
@@ -166,14 +171,76 @@ describe('App', () => {
...
+    }));
+
```
ThomasSchellenbergNextCentury, line 278: I think you can also test that the subscriptions are called when an event is published like this:
```javascript
it('does call bindShowSimpleSearch if showSimpleSearch event is published', async(() => {
  let messenger = new neon.eventing.Messenger();
  let spyOnBingShowSimpleSearch = spyOn(component, 'bindShowSimpleSearch');
  messenger.publish('showSimpleSearch', {
    showSimpleSearch: true
  });
  expect(spyOnBingShowSimpleSearch.calls.count()).toEqual(1);
  expect(spyOnBingShowSimpleSearch.calls.argsFor(0)).toEqual([{
    showSimpleSearch: true
  }]);
});
```

src/app/app.component.ts
```diff
@@ -231,6 +235,18 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
...
+        this.showSimpleSearch = message.showSimpleSearch;
+    }
+
+    bindShowVisShortcut(message) {
+        this.showVisShortcut = message.showVisShortcut;
+    }
```
ThomasSchellenbergNextCentury: You can add tests for the bind* functions like:
```javascript
it('bindShowVisShortcut does update showVisShortcut', async(() => {
  expect(component.showVisShortcut).toEqual(false);
  expect(debugElement.query(By.css('#showVisShortcutButton'))).toBeNull();
  component.bindShowVisShortcut({
    showSimpleSearch: true
  });
  expect(component.showVisShortcut).toEqual(true);
  fixture.detectChanges();
  fixture.whenStable().then(() => {
    fixture.detectChanges();
    expect(debugElement.query(By.css('#showVisShortcutButton'))).not.toBeNull();
  });
});
```
(In the above example, you'd need to add `id=showVisShortcutButton` to the button element in the HTML.)

src/app/app.component.ts
```diff
@@ -417,12 +433,9 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
...
     ngOnInit(): void {
-        this.messenger.subscribe('showSimpleSearch', (message) => {
-            this.showSimpleSearch = message.showSimpleSearch;
-        });
-        this.messenger.subscribe('showVisShortcut', (message) => {
-            this.showVisShortcut = message.showVisShortcut;
-        });
+        this.messenger.subscribe('showSimpleSearch', (message) => this.bindShowSimpleSearch(message));
+        this.messenger.subscribe('showVisShortcut', (message) => this.bindShowVisShortcut(message));
+        this.messenger.subscribe('showFilterBuilderIcon', (message) => this.bindShowFilterBuilderIcon(message));
```
ThomasSchellenbergNextCentury: Rather than passing `message` to the bind* functions, you could pass `message.showWhateverProperty` and the bind* functions could just accept a boolean.

src/app/app.component.ts
```diff
@@ -475,6 +488,18 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
         });
     }
 
+    publishShowSimpleSearch() {
+        this.messenger.publish('showSimpleSearch', {
+            showSimpleSearch: this.showSimpleSearch
+        });
+    }
```
ThomasSchellenbergNextCentury: I don't think showSimpleSearch or publishShowSimpleSearch are used anywhere in the App Component.  I only see them used in the Settings and Simple Filter Components.  Unless I'm mistaken, you should just remove them from here.

src/app/components/add-visualization/add-visualization.component.spec.ts
```diff
@@ -81,13 +93,46 @@ describe('Component: AddVisualization', () => {
...
+
+    it('tests default values', (() => {
+        expect(component.showVisShortcut).toEqual(true);
+        expect(component.selectedIndex).toEqual(-1);
+        expect(component.messenger).toBeTruthy();
+    }));
+
+    it('Check that the publish function updates the correct booleans', (() => {
```
ThomasSchellenbergNextCentury, line 119: The name of this test is odd.  I don't think we're actually publishing anything here...?

src/app/components/settings/settings.component.ts
```diff
@@ -89,9 +90,12 @@ export class SettingsComponent implements OnInit, OnDestroy {
         this.formData.currentTheme = this.widgetService.getTheme();
         this.checkSimpleFilter();
         this.validateDatasetService();
+
+        this.messenger.subscribe('showFilterBuilderIcon', (message) => {
+            this.showFilterBuilderIcon = message.showFilterBuilderIcon;
+        });
```
ThomasSchellenbergNextCentury, line 96: I think if you moved the 4 subscription functions from here into their own functions (like you did with the bind* functions) then we could test them like you tested the other components.

src/app/components/simple-filter/simple-filter.component.spec.ts
```diff
@@ -254,4 +283,31 @@ describe('Component: SimpleFilter unconfigured', () => {
...
+
+        expect(tester.spyOnInit.calls.count()).toEqual(1);
+
+        tester.component.showSimpleSearch = false;
+        expect(tester.component.showSimpleSearch).toEqual(false);
+        tester.component.ngOnInit();
+
+        tester.component.bindShowSimpleSearch(message);
+        tester.component.ngOnInit();
+        tester.component.publishShowSimpleSearch();
```
ThomasSchellenbergNextCentury, line 306: I think this could be separated out into multiple unit tests:  one that checks that the subscription is done on init, and another that checks that the publish function works.

ThomasSchellenbergNextCentury:
Hey @Konnoke I know I added a bunch of comments but I wanted you to know that I think you're making good progress.  Writing simple tests is easy, but writing good tests is hard and takes a lot of practice.  You're definitely learning though!  :)

src/app/app.component.ts
Konnoke: I changed it to getShowSimpleSearch because there was already a boolean named 'showSimpleSearch'

src/app/app.component.ts
Konnoke: I tried something like this before but I would get an issue "Cannot find name 'By'". 

One of the solutions recommended was changing the target/module in the tsconfig.json to es6/es2016 instead of es5/es2015, which is what we have. I decided not to change it at the time and erased the test because changing it seems like it might result in other issues, but I'll try changing it to see if anything breaks.

src/app/app.component.spec.ts
```diff
@@ -149,22 +157,129 @@ describe('App', () => {
...
+            showFilterBuilderIcon: false,
+            showVisShortcut: false
+        };
+
+        expect(spyOnMessengerSubscribe.calls.count()).toEqual(0);
+        expect(spyOnInit.calls.count()).toEqual(1);
+        component.ngOnInit();
+        expect(spyOnInit.calls.count()).toEqual(2);
+        //This fails because the subscribe call count is zero and I'm not sure why it's not three.
+        //expect(spyOnMessengerSubscribe.calls.count()).toEqual(3);
```
ThomasSchellenbergNextCentury: @Konnoke I'm guessing it's zero because of some shenanigans due to the Messenger not being an Angular class.  Since it's buggy and not really doing anything, you should just remove the spyOnMessengerSubscribe.

src/app/app.component.ts
ThomasSchellenbergNextCentury: @Konnoke Hmm, I'd actually rename it to something like "updateShowSimpleSearch", since any function that starts with "get" makes me think it's a getter (and returns something), which isn't the case.

PR #117, opened 2018-12-04
Changed the ColorSets to use CSS Custom Properties

ThomasSchellenbergNextCentury:
I changed the ColorSets (that supply the colors for the charts/points/etc.) to use CSS custom properties rather than hard-coded values.  The custom properties are needed for the colors to update whenever the theme is changed.  I also had to change Color to handle CSS strings rather than RGB values.  I added hex codes to the WidgetService for the main and accent colors for all themes.

Please note that the Dark and Green themes are just copies of the Teal theme for now (there's already a TODO with a JIRA ticket number in them).  Thanks!

### DIFF
--- a/src/app/color.ts
+++ b/src/app/color.ts
@@ -21,11 +21,14 @@ import { ElementRef } from '@angular/core';
  */
 export class Color {
     /**
-     * Creates and returns a Color object using the given Hex string like #123 or #112233.
+     * Creates and returns a Color object using the given Hex string like "#123" or "#112233" or "112233".
      * @arg {string} inputHex
      * @return {Color}
      */
     static fromHexString(inputHex: string): Color {
+        if (!inputHex) {
+            return null;
+        }
         // https://stackoverflow.com/questions/5623838/rgb-to-hex-and-hex-to-rgb
         let shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
         let hex = inputHex.replace(shorthandRegex, function(m, r, g, b) {
@@ -34,7 +37,8 @@ export class Color {
         let hexArray = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
         if (hexArray) {
             let rgb = parseInt(hexArray[1], 16) + ',' + parseInt(hexArray[2], 16) + ',' + parseInt(hexArray[3], 16);
-            return new Color(inputHex, 'rgba(' + rgb + ',0.66)', 'rgba(' + rgb + ',0.33)');
+            return new Color((inputHex.indexOf('#') === 0 ? inputHex : ('#' + inputHex)), 'rgba(' + rgb + ',0.66)',
+                'rgba(' + rgb + ',0.33)');
         }
         return null;
     }
@@ -67,7 +71,7 @@ export class Color {
     }
 
     /**
-     * Creates and returns a Color object using the given RGB string like rgb(12, 34, 56).
+     * Creates and returns a Color object using the given RGB string like "12,34,56".
      * @arg {string} inputRGB
      * @return {Color}
      */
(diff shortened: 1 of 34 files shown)


src/app/color.ts
```diff
@@ -34,151 +31,174 @@ export class Color {
...
-    private getBase255(value: number) {
-        return value;
+    /**
+     * @constructor
+     * @arg {string} css
+     * @arg {string} opacity66
+     * @arg {string} opacity33
+     */
+    constructor(private css: string, private opacity66: string, private opacity33: string) {
+        // Do nothing.
```
sharknc, line 93: It looks like you have hard coded for opacity values into the variable names, method names, and your css. Would it be better to make this more generic so if the values need to be changed in the future you do not have to update the code and css? For example using something like _opacity-high_ and _opacity-medium_ instead of _opacity66_ and _opacity33_ respectively?

src/app/color.ts
```diff
@@ -13,16 +13,13 @@
  * limitations under the License.
```
sharknc, line 13: I could not located a color.spec.ts. I see some of the component tests below cover testing some of the functionality, but I did not see a test case that covers the computeColor method. Would it be worth creating its own set of tests for completeness?

src/app/color.ts
```diff
@@ -34,151 +31,174 @@ export class Color {
...
-        // NEON TEAL COLOR THEME - DARK
-        new Color(49, 39, 97),    // #312761 (deep purple)
-        new Color(198, 88, 0),    // #C65800 (orange)
-        new Color(129, 29, 100),  // #811D64 (purple)
-        new Color(126, 146, 0),   // #7E9200 (lime)
-        new Color(187, 30, 69),   // #BB1E45 (pink)
-        new Color(0, 108, 203),   // #006CCB (blue)
-        new Color(199, 165, 0),   // #C7A500 (yellow)
-        new Color(54, 154, 82)    // #369A52 (sea green)
+        new Color('var(--color-set-1)', 'var(--color-set-1-opacity-66)', 'var(--color-set-1-opacity-33)'),
```
sharknc: Do we need to have private opacity66: string, private opacity33 as part of the constructor for Color? It looks like you only really need the first parameter _css_ to get the rest of the information you need. For example passing _var(--color-set-1)_ the _--color-set-1_ is the prefix to opacity-66 and opacity-33.

src/app/color.ts
```diff
@@ -34,151 +31,174 @@ export class Color {
...
 
     /**
-     * Get the 'inactive' color, aka the RGBA string with an alpha of 0.3
+     * Returns the CSS for the color with opacity 33 using the given ElementRef object to find custom CSS properties like "--variable".
+     * @arg {ElementRef} elementRef
      * @return {string}
      */
-    getInactiveRgba(): string {
-        return this.toRgba(0.2);
+    public getComputedCssOpacity33(elementRef: ElementRef): string {
```
sharknc: The methods _getComputedCss_, _getComputedCssOpacity66_, and _getComputedCssOpacity33_ seems redundant. It feels like this could be reduced down to a single method _getComputedCss_. I am okay with leaving it, just wanted to make a note in case you agree.

src/app/color.ts
ThomasSchellenbergNextCentury: Yes!  I'll try to do them or at least add a ticket for it.

src/app/color.ts
ThomasSchellenbergNextCentury: Yeah, that's probably a good idea.

src/app/color.ts
ThomasSchellenbergNextCentury: So, I think that would introduce an assumption that might not be completely obvious to future users of Color.  If I created a Color, I'd have to know that it automatically added a suffix onto the end of the constructor argument and name my custom CSS properties accordingly.  I think the current way is slightly clearer, if a little more verbose.  That being said, I think either approach is fine, and they each have their pros and cons.

src/app/color.ts
ThomasSchellenbergNextCentury: Either you have 3 accessors for the 3 different properties or you have 1 accessor that takes an argument and returns 1 of 3 properties based on that argument.  I think either would be fine, but I decided to use the 1-accessor-per-property approach because that seemed clearer to me.

src/app/color.ts
sharknc: Okay sounds good. I have no issue with leaving it.

src/app/color.ts
sharknc: Okay, sounds good.

sharknc (dismissed):
- All tests ran successfully
- Color service worked as expected when running application.

Konnoke (dismissed):

lbraun2 (dismissed):

Konnoke:
I have had 13 unit tests that failed for me, all of them are from the aggregation component.
example: Component: Aggregation onQuerySuccess with XY data does update expected properties and call expected functions FAILED

sharknc:
> I have had 13 unit tests that failed for me, all of them are from the aggregation component.
> example: Component: Aggregation onQuerySuccess with XY data does update expected properties and call expected functions FAILED

I was unable to replicate these failures. 1944 tests pass and 0 fail for me.

ThomasSchellenbergNextCentury:
@Konnoke I can replicate the failures.  They were caused by my latest commit ("CR updates") and have now been fixed.

I'm guessing that @sharknc couldn't replicate them because he hadn't pulled my "CR updates" commit.

PR #118, opened 2018-12-04
THOR-970 THOR-972 Updates to Map cluster behavior

ThomasSchellenbergNextCentury:
I updated the Map cluster to use the theme colors.  I also changed clustering so it's always on (and can't be disabled) with permission from Sophie.  I did this for two main reasons:  1. Loading a lot of individual points (thousands or more) slows down the browser significantly; 2. Individual points that are very close together (or overlapping) can't be clicked without the cluster behavior.  Finally, I fixed a few bugs with the hover popup behavior.

### DIFF
--- a/src/app/components/map/map.component.html
+++ b/src/app/components/map/map.component.html
@@ -135,14 +135,6 @@
                             </mat-form-field>
                             <p></p>
                         </div>
-
-                        <mat-form-field>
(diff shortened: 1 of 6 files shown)


ScottWarner:
Looks good to me

ScottWarner (approved):

PR #121, opened 2018-12-10
THOR-874/THOR-826

rartiss55:
New multi-select dashboards (linked with THOR-826).


src/app/app.component.html
```diff
@@ -75,11 +72,19 @@
...
         </div>
     </div>
 
+    <div class="header-toggle-container">
+        <div class="dashboard-selector-container" id="dashboard.selector">
+            <app-dashboard-selector (activeDatasetChanged)="simpleFilter.setSimpleFilter(); toggleDashboardSelectorDialog(false)">
+            </app-dashboard-selector>
+            <div class="header-toggle-footer"></div>
+        </div>
+    </div>
```
ThomasSchellenbergNextCentury, line 86: We should make sure to coordinate with @Konnoke because I believe he's touching the same code as part of his latest endeavor with the widget options menu.

src/app/app.component.ts
```diff
@@ -163,6 +164,11 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
             this.domSanitizer.bypassSecurityTrustResourceUrl('../assets/icons/filter_builder_active.svg')
         );
 
+        this.matIconRegistry.addSvgIcon(
+            'dashboard_selector',
+            this.domSanitizer.bypassSecurityTrustResourceUrl('../assets/icons/database_icon.svg')
+        );
```
ThomasSchellenbergNextCentury, line 170: 1. Does this icon actually exist?  I can't find it in src/assets/icons/
2. Should we be using a Material Design icon here instead?

src/app/components/dashboard-dropdown/dashboard-dropdown.component.html
```diff
@@ -0,0 +1,12 @@
...
+        <mat-option *ngFor="let dashboardKey of getDashboardKeys()" [value]="dashboards.choices[dashboardKey]">
+            {{ getDashboardName(dashboardKey) }}
+        </mat-option>
+    </mat-select>
+</mat-form-field>
+
+<app-dashboard-dropdown *ngIf="hasMoreChoices()" #nextDropdown
+    [dashboards]="selectedDashboard"
+    (selectionChange)="onSelectionChange($event)">
+</app-dashboard-dropdown>
```
ThomasSchellenbergNextCentury, line 12: Oh, is this calling itself recursively?  Neat!

src/app/components/dashboard-dropdown/dashboard-dropdown.component.ts
```diff
@@ -0,0 +1,129 @@
...
+     */
+    emitSelectedDashboard() {
+        this.changeDetection.detectChanges();
+
+        // if no more choices/dropdowns exist, emit the chosen dashboard
+        // otherwise, emit undefined
+        if (!this.hasMoreChoices()) {
+            this.selectionChange.emit(this.selectedDashboard);
+        } else if (this.hasMoreChoices() && this.selectedDashboard) {
+            this.selectionChange.emit();
```
ThomasSchellenbergNextCentury, line 74: What's the point of emitting undefined here?

src/app/components/dashboard-dropdown/dashboard-dropdown.component.ts
```diff
@@ -0,0 +1,129 @@
...
+            _.findKey(this.dashboards.choices, <any> this.selectedDashboard)
+        );
+    }
+
+    /**
+     * Propagates selection back up to parent component.
+     *
+     * @arg {any} $event
+     */
+    onSelectionChange($event) {
```
ThomasSchellenbergNextCentury: So onSelectionChange is only called when the child emits a change event, right?  I think it's a bit confusing since this element also has its own selection element that emits its own change events, but they aren't handled with onSelectionChange.  Maybe we could rename this function to something like "onChildSelectionChange" or "onNestedSelectionChange"?

src/app/components/dashboard-dropdown/dashboard-dropdown.component.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+import { DashboardDropdownComponent } from './dashboard-dropdown.component';
+import { DashboardOptions, Dashboard } from '../../dataset';
+
+describe('Component: DashboardDropdown with input', () => {
+    let fixture: ComponentFixture<DashboardDropdownComponent>;
+    let component: DashboardDropdownComponent;
+
+    /* tslint:disable:no-string-literal */
+    let dashboardTableKeys1 = new Map<string, string>();
+    dashboardTableKeys1['tableKey'] = 'datastore1.database1.table1';
```
ThomasSchellenbergNextCentury: Oh, is this a shortcut for Map.set?  Cool.

src/app/components/dashboard-dropdown/dashboard-dropdown.component.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+
+        component.emitSelectedDashboard();
+        expect(spy.calls.count()).toEqual(1);
+    }));
+
+    it('emitSelectedDashboard() should emit selectedDashboard if no more choices exists', (() => {
+        let spy = spyOn(component.selectionChange, 'emit');
+
+        /* tslint:disable:no-string-literal */
+        component.selectedDashboard = component.dashboards.choices['dash1'];
```
ThomasSchellenbergNextCentury: Huh, all these tslint commands are annoying.  Would it be easier to just use get()?  I guess ideally we could customize the no-string-literal rule to ignore Maps but it doesn't look like that's an option.

src/app/components/dashboard-dropdown/dashboard-dropdown.component.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+    }));
+});
+
+describe('Component: DashboardDropdown with no inputs', () => {
+    let fixture: ComponentFixture<DashboardDropdownComponent>;
+    let component: DashboardDropdownComponent;
+
+    /* tslint:disable:no-string-literal */
+    let dashboardTableKeys1 = new Map<string, string>();
+    dashboardTableKeys1['tableKey'] = 'datastore1.database1.table1';
```
ThomasSchellenbergNextCentury: Are these all the same variables that were defined in the describe() above?  Could we move them outside the scope of describe() so we don't need to repeat them in each one?

src/app/components/dashboard-dropdown/dashboard-dropdown.component.spec.ts
eaciv: You guys may want to double check this. I think this is simply setting the 'tableKey' property on the dashboardTableKeys1 object, not really using the map interface. If that's what you want, there's no reason to change, but if you really want the map functionality, you may want to use Map.set.

src/app/components/dashboard-selector/dashboard-selector.component.ts
```diff
@@ -131,77 +135,95 @@ export class DatasetSelectorComponent implements OnInit, OnDestroy {
...
-                let index = this.findMatchingIndex(configItem);
-                let dataset = this.datasets[index];
+            let dashboardKeys = this.dashboards.choices ? Object.keys(this.dashboards.choices) : [];
 
-                if (configItem.options.connectOnLoad) {
-                    this.connectOnLoad = true;
-                }
+            if (this.hasConnectOnLoadDashboard(this.dashboards.choices, dashboardKeys)) {
+                this.connectOnLoad = true;
+                let paths = this.connectOnLoadDashboard.pathFromTop.replace(/choices./, '').split(/(?:.choices.)/);
```
ThomasSchellenbergNextCentury: Do we really need "choices" to separate the different choices?  Couldn't we just split on a period?

Also, what happens if "choices." is part of the ID of one of the choices (unlikely though that may be)?

src/app/components/dashboard-selector/dashboard-selector.component.ts
```diff
@@ -131,77 +135,95 @@ export class DatasetSelectorComponent implements OnInit, OnDestroy {
...
+     * @param {string[]} keys
+     * @return {Boolean}
+     */
+    hasConnectOnLoadDashboard(dashboardChoices: Map<string, Dashboard>, keys: string[]): boolean {
+        return keys.some((choiceKey) => {
+            let nestedChoiceKeys = dashboardChoices[choiceKey].choices ? Object.keys(dashboardChoices[choiceKey].choices) : [];
+
+            if (!nestedChoiceKeys.length) {
+                if (dashboardChoices[choiceKey].options && dashboardChoices[choiceKey].options.connectOnLoad === true) {
+                    this.connectOnLoadDashboard = dashboardChoices[choiceKey];
```
ThomasSchellenbergNextCentury: Do we need to save connectOnLoadDashboard as a class variable?  It looks like it's only used in one place.  Couldn't we just return it as part of this function?

src/app/components/dashboard-dropdown/dashboard-dropdown.component.spec.ts
ThomasSchellenbergNextCentury: Ah, I was wondering why I couldn't find any documentation on this shorthand.  Thanks @ecurley-nextcentury !

src/app/components/dashboard-dropdown/dashboard-dropdown.component.ts
rartiss55: I think this was to handle the use case where a user changes a previous dropdown choice once making a final dashboard selection -- needed a way to reset it and send it up the chain.

ThomasSchellenbergNextCentury: Cool!  Thanks.

src/app/components/dashboard-dropdown/dashboard-dropdown.component.spec.ts
rartiss55: Changing this from Map to Object after talking with Thomas.

lbraun2 (approved):

src/app/app.component.ts
rartiss55: I added it -- the icon we originally wanted to use from Material Design is being used for the custom database connection. I think later when things are more combined we'll be able to get rid of this one I added.

src/app/components/dashboard-selector/dashboard-selector.component.ts
rartiss55: It seemed less weird to me to account for choices within the dropdown component rather than here, but to handle the use case you're talking about, we could split on a period, and do something like this in the dashboard dropdown selectDashboardChoice function:

        let choicePath = paths[indexToUse];
        let keyWithinChoice = paths[indexToUse + 1];
        dropdown.selectedDashboard = dashboard[choicePath][keyWithinChoice];

As opposed to what it is now:
         let path = paths[indexToUse];
         dropdown.selectedDashboard = dashboard.choices[path];

If you'd prefer? Then again there's still the use case where someone puts a period in a choice key, and not sure what to do then. Wondering if the way to handle this is to reject  or replace some of these values in validateDashboards instead?

src/app/components/dashboard-dropdown/dashboard-dropdown.component.spec.ts
rartiss55: Took these out.

src/app/components/dashboard-dropdown/dashboard-dropdown.component.spec.ts
rartiss55: Done!

src/app/components/dashboard-selector/dashboard-selector.component.ts
ThomasSchellenbergNextCentury: I'm confused.  I thought I was talking about the "choices" strings in the "pathFromTop" variable, and it seems like you're talking about the "choices" object in the Dashboard class?  Aren't they two separate things?

ThomasSchellenbergNextCentury: @rartiss55 OK, new idea: Can we just change pathFromTop to a string array?  I think it will be easy to change (although I'm sure you'd know better!):

Defining pathFromTop:  Concat onto the array
https://github.com/NextCenturyCorporation/neon-dash-internal/pull/121/files#diff-d169b3ccf465ae50dc6eb2077f42acceR139

Checking connectOnLoad:  Already split into an array!
https://github.com/NextCenturyCorporation/neon-dash-internal/pull/121/files#diff-c534309f2df64e3786db3bcec17c3423R202

Connecting to preset (inside _.get):  Use join('.choices.')
https://github.com/NextCenturyCorporation/neon-dash-internal/pull/121/files#diff-c534309f2df64e3786db3bcec17c3423R319

src/app/components/dashboard-selector/dashboard-selector.component.ts
ThomasSchellenbergNextCentury: @rartiss55 Right now, you're actually returning `dashboardChoices[choiceKey]`, which is a Dashboard object.  I'm assuming it's getting converted into a boolean.  I think you can just change the return type to Dashboard.  Remember to update the docs too.

ThomasSchellenbergNextCentury: @rartiss55 You can change `some` to `reduce`:

```javascript
        let connectOnLoadDashboards = keys.reduce((dashboards, choiceKey) => { // changed
            let nestedChoiceKeys = dashboardChoices[choiceKey].choices ? Object.keys(dashboardChoices[choiceKey].choices) : [];

            if (!nestedChoiceKeys.length) {
                if (dashboardChoices[choiceKey].options && dashboardChoices[choiceKey].options.connectOnLoad === true) {
                    dashboards.push(dashboardChoices[choiceKey]); // changed
                }
                return dashboards; // added
            } else {
                return dashboards.concat(this.hasConnectOnLoadDashboard(dashboardChoices[choiceKey].choices, nestedChoiceKeys)); // changed
            }
        }, []);
        return connectOnLoadDashboards.length ? connectOnLoadDashboards[0] : null; // added
```

src/app/components/dashboard-dropdown/dashboard-dropdown.component.ts
rartiss55: Went with "onChildSelectionChange"

src/app/components/dashboard-selector/dashboard-selector.component.ts
ThomasSchellenbergNextCentury: @rartiss55 

```javascript
for (let choiceKey in keys) {
    let nestedChoiceKeys = dashboardChoices[choiceKey].choices ? Object.keys(dashboardChoices[choiceKey].choices) : [];
    if (!nestedChoiceKeys.length) {
        if (dashboardChoices[choiceKey].options && dashboardChoices[choiceKey].options.connectOnLoad === true) {
            return dashboardChoices[choiceKey];
        }
    } else {
        let nestedDashboard = this.hasConnectOnLoadDashboard(dashboardChoices[choiceKey].choices, nestedChoiceKeys);
        if (nestedDashboard) {
            return nestedDashboard;
        }
    }
}
return null;
```

src/app/app.component.html
Konnoke: Yea there will be conflicts with my branch but it'll be fine. 

Instead of having two header-toggle-container, it might be better to have it underneath the first header-toggle-container.

In my branch I fixed THOR-983 which involved creating a create boolean for each element such as (the gear, add vis, settings etc) so the component are only created once and togging it only changes the display style attribute to none/show

src/app/components/dashboard-selector/dashboard-selector.component.ts
rartiss55: Done!

src/app/components/dashboard-selector/dashboard-selector.component.ts
rartiss55: Added.

rartiss55:
@ThomasSchellenbergNextCentury I think everything got addressed but let me know if there was anything else. Thanks!

PR #123, opened 2018-12-13
THOR-549 Combined BaseNeonComponent and BaseLayeredNeonComponent Superclasses

ThomasSchellenbergNextCentury:
Combined the BaseNeonComponent and the BaseLayeredNeonComponent.  The main changes are in the BaseNeonComponent.

BaseNeonComponent has an "options" property which has an array of "layers".  Widgets can ignore the "layers" and use the "options" directly if they only need bindings for a single layer.  Otherwise they can create and use their "layers" which are the same type of objects as the top-level "options" and can thus be used in the same way.

Accomplishments:

I added the layer functions (like addLayer, postAddLayer, and removeLayer) to BaseNeonComponent.  Each function takes a WidgetOptionCollection which is probably one of the "layers".

I added an "options" (WidgetOptionCollection) argument to many of the other functions (filtering, searching, config-change-handling, etc.) and updated the functions to use the input "options" rather than "this.options".  The input "options" will either be "this.options" or a layer from "layers" so thus they are now useable by both single-layer and multi-layer widgets.

I updated some existing properties in BaseNeonComponent (like "isLoading" and "outstandingDataQueriesByLayer") and added a new property "isMultiLayerWidget".

I also started to do some cleanup in BaseNeonComponent but there was a lot of it that I decided to move to a separate PR.

I know there's a lot going on here so please don't hesitate to ask any questions.  Thanks for the help!


src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -139,32 +148,70 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
     abstract subNgOnDestroy();
 
     /**
-     * Returns the export header data.
+     * Adds a new layer for the visualization using the given bindings.
+     *
+     * @arg {any} [options]
```
ThomasSchellenbergNextCentury, line 153: This documentation needs to say "A WidgetOptionCollection object"

src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -703,13 +789,38 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
...
+
+        if (this.options.layers.length) {
+            return this.options.layers.map((layer) => {
+                let text = this.createButtonText(layer, this.options.limit);
+                return text ? (layer.title + ' (' + text + ')') : '';
+            }).filter((text) => !!text).join(', ');
+        }
+
+        return '';
+    }
```
ThomasSchellenbergNextCentury, line 814: This was all just taken from BaseLayeredNeonComponent.

Konnoke (approved):

bdpham (approved):

PR #124, opened 2018-12-13
Aida 466

bdpham:
Cherry-picked stable changes from verdi-demo branch.
Notable changes:
- Datatable ignoreself option now works with multi-filters
- Colorsets now have globalcolor mapping for specified fields (config colormaps)
- Thumbnail grid now has baseball card view
- Taxonomy viewer now shows node count

@ecurley-nextcentury @snorpet Can you test with your respective configs and data?


src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -495,10 +498,39 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
+    onEvent = ($event) => {
         //Intentionally empty
     }
 
+    addCountsToTaxonomy(data: any, groups: any[]) {
+        for (let group of groups) {
+            let count = 0;
+            group.nodeIds = [];
+            data.forEach((d) => {
+                let id = neonUtilities.deepFind(d, this.options.idField.columnName);
```
gitnordt, line 542: I believe the sourceIdField code needs to bve added since it is used to add the node counts. Otherwise, this function is just counting the categories and types.

gitnordt (changes requested):

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
bdpham: I'm not quite sure what the sourceField code is and where I can find it. I cherry picked this from the AIDA-416 PR. Can you point me to where I need to look?

eaciv:
As I only had minor changes to this branch, I don't feel comfortable approving it. However, this branch worked for my TA1 layout. Let me know if you need anything further from me.

src/app/components/visualization-injector/visualization-injector.component.ts
```diff
@@ -37,7 +37,11 @@ import { VisualizationService } from '../../services/visualization.service';
 import { ThumbnailGridComponent } from '../thumbnail-grid/thumbnail-grid.component';
 import { NewsFeedComponent } from '../news-feed/news-feed.component';
 import { QueryBarComponent } from '../query-bar/query-bar.component';
-import { ThumbnailDetailsContractedComponent, ThumbnailDetailsExpandedComponent } from '../thumbnail-grid/thumbnail-details.component';
+import {
+    CardThumbnailSubComponent,
+    DetailsThumbnailSubComponent,
+    TitleThumbnailSubComponent
+} from '../thumbnail-grid/thumbnail-details.component';
```
ThomasSchellenbergNextCentury: FYI You don't actually need to import these components here since they're not being used directly by the visualization-injector.

src/app/components/visualization-injector/visualization-injector.component.spec.ts
```diff
@@ -48,7 +48,10 @@ import { ThumbnailGridComponent } from '../thumbnail-grid/thumbnail-grid.compone
 import { initializeTestBed } from '../../../testUtils/initializeTestBed';
 import { NewsFeedComponent } from '../news-feed/news-feed.component';
 import { QueryBarComponent } from '../query-bar/query-bar.component';
-import { ThumbnailDetailsContractedComponent, ThumbnailDetailsExpandedComponent } from '../thumbnail-grid/thumbnail-details.component';
+import { TitleThumbnailSubComponent,
+    DetailsThumbnailSubComponent,
+    CardThumbnailSubComponent
+} from '../thumbnail-grid/thumbnail-details.component';
```
ThomasSchellenbergNextCentury: FYI You don't actually need to import these components here since they're not being used directly by the visualization-injector.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -50,10 +50,13 @@ export class TaxonomyViewerOptions extends BaseNeonOptions {
     public id: string;
     public categoryField: FieldMetaData;
     public idField: FieldMetaData;
+    public sourceIdField: FieldMetaData;
     public typeField: FieldMetaData;
     public subTypeField: FieldMetaData;
     public filterFields: string[];
     public ignoreSelf: boolean;
+    public extendedFilter: boolean;
+    public extensionFields: any[];
```
ThomasSchellenbergNextCentury: The extensionFields aren't actually used anywhere.  If there's really not a use for them, you can just delete them.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -466,7 +501,9 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
 
                 });
 
+                this.addCountsToTaxonomy(response.data, this.taxonomyGroups);
                 this.sortTaxonomyArrays(this.taxonomyGroups);
+                //console.log(this.taxonomyGroups);
```
ThomasSchellenbergNextCentury: Remove unneeded comment

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -668,10 +750,27 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
      * Called after the filters in the filter service have changed.
      * Defaults to calling setupFilters() then executeQueryChain()
      */
-    handleFiltersChangedEvent(): void {
-        //TODO:Update Taxonomy when filter from filter tray is deleted AIDA-390
-        /*        console.log(this.getElementRefs().treeRoot.treeModel.nodes);
-                console.log(this.treeRoot.treeModel.nodes);*/
+    handleFiltersChangedEvent($event): void {
```
ThomasSchellenbergNextCentury, line 750: So what's the goal of overriding this function?  Is it to update the visualization without requerying?

bdpham: @snorpet Can you help answer this question?

src/app/components/thumbnail-grid/thumbnail-details.component.ts
```diff
@@ -18,12 +17,30 @@ import { Component, Input } from '@angular/core';
...
 
 @Component({
-    selector: 'app-thumbnail-details-contracted',
-    templateUrl: './thumbnail-details-contracted.component.html',
+    selector: 'app-subcomponent-title-thumbnail',
+    templateUrl: './subcomponent.title-view.html',
+    styleUrls: ['./thumbnail-grid.component.scss']
+})
+
+export class TitleThumbnailSubComponent {
```
ThomasSchellenbergNextCentury: Can we move the 3 subcomponents into their own separate files please?  If we've done the work to create 3 distinct classes, that indicates that we want the subcomponents to be able to exist independently, but they can't do it properly if they're all in the same file together!

ThomasSchellenbergNextCentury:
@bdpham I'm getting errors in the TaxonomyViewer tests for createQuery and getExportFields.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
gitnordt: @ThomasSchellenbergNextCentury  @bdpham I am using the handleFiltersChangedEvent to handle the filter changes in the taxonomy component. Requering is still occurring after my defined actions are executed. This is a special case so if you want to discuss in more detail, we can take it offline. I am simply using a method already being executed instead a creating a new method with repetitive functionality. A notable difference is that I changed the base-neon.component version to accept an event argument when necessary. I believe this will be a useful change and the argument is not required so it shouldn't affect other code.

ThomasSchellenbergNextCentury: @snorpet Where are you requerying?  The original handleFiltersChangedEvent called executeQueryChain, but you're overriding that behavior.  ~~The only time this would trigger a query is if you call removeFilter~~, but it looks like that's not guaranteed to happen.  Edit:  I was wrong:  removeFilter is not querying at all.

src/app/components/thumbnail-grid/thumbnail-details.component.ts
gitnordt: This statement is inaccurate. These are meant to be sub-components of the thumbnail grid component. If the desire is to have these as independent components, putting their classes in separate files are not going to accomplish this. All components would have to be re-worked to accomplish the thumbnail grid functionality and styling. Afterwards, the thumbnail grid component would be deprecated because it wouldn't have a use anymore. In this case, the classes being housed in the same file does not have an impact on usability. This change request is personal preference, no exclamation needed.

ThomasSchellenbergNextCentury: @snorpet If they're all dependent on the thumbnail-grid, why bother separating them into 3 classes?  Why not just have 1 class with an ngIf on all 3 templates?  The main reason for having separate classes is if there's a chance we're going to reuse them independently (for example, import the Card into another widget, but not the Details or the Title).

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
gitnordt: executeQueryChain is still being called after my code is executed. As I suggested before, we can take the conversation offline. It is not being overriden. If you are expecting JAVA-like override behavior, you may be disappointed. JavaScript override works a little differently due to inheritance from my understanding.

src/app/components/thumbnail-grid/thumbnail-details.component.ts
gitnordt: That is fine. If you would like me to change them to templates, please put in a ticket and assign it to me. This should not hold up Bao's PR as this is already code that was previously approved by you and pushed into the AIDA master branch.

gitnordt (changes requested):
Bao, I noticed that one change is still missing from the taxonomyviewercomponent.html, line 122. This is the line that allows for the nodeCount to be displayed in the taxonomy.

<span (click)="onEvent($event, node)">{{ node.data.name }} <span *ngIf="node.data.nodeCount > 0">({{ node.data.nodeCount }})</span></span>

src/app/components/thumbnail-grid/thumbnail-details.component.ts
ThomasSchellenbergNextCentury: @bdpham Please add a (low priority) ticket for this.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
ThomasSchellenbergNextCentury: @bdpham and I have held a discussion and he will be adding a ticket for any needed bug fixing.

ThomasSchellenbergNextCentury (approved):

gitnordt (approved):

PR #125, opened 2018-12-20
THOR-943

rartiss55:
- Combining filter tray and filter builder.
- Removing filter builder as a visualization.


wagnerwj:
Nag Thomas about this one, Rachel. I'd like to see this sooner.

Konnoke (commented):
What are the advantages of having the filter component handle the filter builder and current filter components vs having the app component handle it?

src/app/components/filters/filters.component.ts
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+import { Component, EventEmitter, Input, Output } from '@angular/core';
+
+import { BaseNeonComponent } from '../base-neon-component/base-neon.component';
+import { BaseLayeredNeonComponent } from '../base-neon-component/base-layered-neon.component';
+
+import * as _ from 'lodash';
```
Konnoke: I don't see lodash being used in the filter component.

rartiss55:
@Konnoke I thought it was better for code maintenance and testability to have a separate component
manage the filters rather than adding it to the app component. Plus it would make it easier if both filter components were further combined later.

ThomasSchellenbergNextCentury (dismissed):

Konnoke:
Expected behavior, but old save states that have the filter builder component in the dashboard will still load but the filter builder component in the dashboard will be blank (visualization container component with nothing in it)

rartiss55:
@Konnoke I think that's expected because we're phasing out the filter builder as a visualization. @ThomasSchellenbergNextCentury does this sound fine to you?

Konnoke (dismissed):

Konnoke (dismissed):

src/app/components/filters/filters.component.ts
rartiss55: you're right -- removed.

ThomasSchellenbergNextCentury:
@rartiss55 @Konnoke 

I think a saved state loading a blank widget in place of the filter builder is fine.  If it becomes an issue later then we can probably implement something to detect it and ignore it.

ThomasSchellenbergNextCentury (dismissed):

rartiss55:
@ThomasSchellenbergNextCentury merged in lorelei-master but the way I did it must have cancelled out approvals. Please re-approve when you get the chance. Thanks!

ThomasSchellenbergNextCentury (approved):

PR #126, opened 2018-12-20
THOR-971 Refactor BaseNeonComponent

ThomasSchellenbergNextCentury:
Please review the refactoring in BaseNeonComponent (and, if you have the time, its subclasses).

My goals were to move shared code from the individual subclasses into the BaseNeonComponent superclass.  This includes:
- Creating queries
- Handling post-query behavior and transforming query results
- Running count queries
- Saving transformed query results and counts
- Pagination

Major changes include:
- Repacling createQuery with finalizeVisualizationQuery
- Replacing onQuerySuccess with transformQueryResults
- Saving the "active data" in the superclass rather than the subclass (except in the annotation-viewer, media-viewer, network-graph, text-cloud, thumbnail-grid, and timeline, which are all marked with "TODO THOR-985")

I also simplified the widget initialization behavior, wrote BaseNeonComponent unit tests, and did other cleanup.  Note that I'll work on refactoring the filter functions next month.

Please let me know if you have any questions.  Thanks for the help!


src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -574,68 +826,82 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
...
+     * Updates tables, fields, and filters whenenver the database is changed and then runs the visualization query.
      *
      * @arg {any} options A WidgetOptionCollection object.
      */
-    handleChangeDatabase(options: any) {
+    public handleChangeDatabase(options: any): void {
         this.updateTablesInOptions(options);
-        this.initializeFieldsInOptions(options, this.createLayerFieldOptions());
+        // TODO FIXME How to do this without the direct comparison
+        if (options === this.options) {
```
rartiss55, line 838: Why does one of these need an unsharedFilterField and the other does not?

rartiss55:
I'm seeing an issue with the data table and annotation viewer -- I attached a screenshot to the THOR-971 ticket for reference.

rartiss55:
@ThomasSchellenbergNextCentury looked into the issue I mentioned (at least for the data-table) -- it looks like before we were doing a query for all fields but now its only querying for fields of optionType FIELD or FIELD_ARRAY. In the case of the data table, we also need the columns displayed, which in my case are part of exceptionsToStatus (which is a NON_PRIMITIVE type). Maybe we can add an additional "fieldsToDisplay" FIELD_ARRAY, create it using the other options mentioned, or move exceptionsToStatus to field options? 

This might require some more thought, and its also worth checking each visualizations again just to see if there's other ones affected by this change.

src/app/components/base-neon-component/base-neon.component.ts
ThomasSchellenbergNextCentury: We need separate behavior depending on whether the given options are the top-level options or layer options.  Note that one part is calling createFieldOptions and the other is calling createLayerFieldOptions.  Also, the top-level options always have an unsharedFilterField, in addition to visualization-specific fields.  I've added a comment.

ThomasSchellenbergNextCentury:
@rartiss55 I fixed the field issue in the data table and annotation viewer:  https://github.com/NextCenturyCorporation/neon-dash-internal/pull/126/commits/1114206d49988b9a8a2f23dc84f9502b95dc1a4a

rartiss55 (approved):

PR #130, opened 2018-12-27
Created a separate config file for the simplified seedling data

gitnordt:
This config file will include a layout for multiple TA3 views, but currently only contains the TA2 and TA3 Network Layouts

### DIFF
--- a/src/app/config/config.verdi-TA2TA3.json
+++ b/src/app/config/config.verdi-TA2TA3.json
@@ -510,11 +510,7 @@
                     "granularity": "day",
                     "filterFields": ["id"],
                     "showOnlyFiltered": true
-                },
-                "minPixelx": 320,
-                "minPixely": 240,
-                "minSizex": 6,
-                "minSizey": 3
+                }
             }
         ]
     }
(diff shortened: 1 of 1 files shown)


src/app/config/config.verdi-TA2TA3.json
```diff
@@ -0,0 +1,521 @@
...
+                    "cleanLegendLabels": true,
+                    "legendFiltering": false
+                },
+                "row": 2,
+                "col": 4
+            },
+            {
+                "name": "Thumbnail Grid",
+                "type": "thumbnailGrid",
+                "icon": "ThumbnailGrid",
```
bdpham: name, icon, minpixel, and minsize are not necessary. There are a few places where they exist, and if they don't serve a purpose, I suggest that they be removed.

gitnordt: I think I've removed all of the fields that were of concern. Please review and let me know if there is anything else. Thanks!

ThomasSchellenbergNextCentury (approved):

PR #136, opened 2019-01-08
THOR-986 Creation of SearchService (Also THOR-876)

ThomasSchellenbergNextCentury:
In this change, I made a new service, SearchService, and pulled the creation of neon.query.Query objects from the individual widgets into this service.  My goal is to compartmentalize the neon.query library to ease a future conversion away from neon.query.Query (for example, to GraphQL).  My next step is to move the creation of neon.query.Filter objects out of the widgets into a service (but that's not included here).  I also fixed THOR-876, allowing the ConnectionService to connect to multiple datastores, though we don't really make use of that capability yet.

Please take a look at AbstractSearchService, SearchService, and my changes to BaseNeonComponent and the finalizeVisualizationQuery functions.

Eric, I think the SearchService class and tests will be especially useful to you.


src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -502,18 +495,19 @@ export abstract class BaseNeonComponent implements AfterViewInit, OnInit, OnDest
...
-                    query.offset((this.page - 1) * this.options.limit);
+                    this.searchService.updateOffset(query, (this.page - 1) * this.options.limit);
                 }
 
                 let filtersToIgnore = this.getFiltersToIgnore();
-                if (filtersToIgnore && filtersToIgnore.length) {
-                    query.ignoreFilters(filtersToIgnore);
+                if (filtersToIgnore && filtersToIgnore.length && (query as any).query) {
+                    // TODO THOR-946
+                    (query as any).query.ignoreFilters(filtersToIgnore);
```
ThomasSchellenbergNextCentury, line 514: We shouldn't be assuming that the NeonQueryPayload has a "query" property (here, and on line 666), and that will get fixed when I work on the filter changes.

src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -1194,52 +1182,6 @@ export abstract class BaseNeonComponent implements AfterViewInit, OnInit, OnDest
...
 
-    /**
-     * Converts data elements that are specified to have a pretty name back to their original data label so that the database can
-     * correctly query. When it comes back in handleSuccessfulVisualizationQuery, the response will be converted back to their "pretty" form
-     * Will probably skew the data if the config specifies a data label to have a pretty name that is the same as another data label
-     *
-     * @arg {any} options A WidgetOptionCollection object.
-     * @arg {neon.query.WherePredicate} wherePredicate
-     */
-    private datifyWherePredicates(options: any, wherePredicate: neon.query.WherePredicate) {
```
ThomasSchellenbergNextCentury, line 1226: This got moved into the SearchService.

src/app/app.component.spec.ts
```diff
@@ -87,68 +90,69 @@ describe('App', () => {
     let component: AppComponent;
     let spyOnInit;
 
-    beforeEach(() => {
```
ThomasSchellenbergNextCentury, line 90: The changes to the tests here, in annotation-viewer, and in simple-filter to use initializeTestBed were unrelated to the main goal of this PR.

src/testUtils/MockServices/SearchServiceMock.ts
```diff
@@ -0,0 +1,171 @@
...
+    NeonFilterClause,
+    NeonQueryGroup,
+    NeonQueryPayload,
+    SortOrder,
+    TimeInterval
+} from '../../app/services/abstract.search.service';
+import { NeonRequest } from '../../app/connection';
+
+/**
+ * Saves filter clauses and query payloads as JSON objects.
```
ThomasSchellenbergNextCentury, line 29: The SearchServiceMock, used by all the unit tests, represents the queries as simple JSON objects to make testing easier and separate them from the neon.query library.

src/app/services/connection.service.ts
```diff
@@ -19,33 +19,28 @@ import * as neon from 'neon-framework';
 @Injectable()
 export class ConnectionService {
 
-    private activeConnection: neon.query.Connection;
+    // Maps the datastore types to datastore hosts to connections.
+    private connections: Map<string, Map<string, neon.query.Connection>> = new Map<string, Map<string, neon.query.Connection>>();
```
rartiss55, line 23: Oh nice -- I didn't account for same database type with different hosts when I did this on the config branch, so I'll make a note :)

src/app/services/abstract.search.service.ts
```diff
@@ -0,0 +1,237 @@
...
+     * Sets the aggregation data on the given search query payload.
+     *
+     * @arg {NeonQueryPayload} queryPayload
+     * @arg {AggregationType} type
+     * @arg {string} name
+     * @arg {string} field
+     * @return {AbstractSearchService}
+     * @abstract
+     */
+    public abstract updateAggregation(queryPayload: NeonQueryPayload, type: AggregationType, name: string,
```
rartiss55, line 165: Why are these functions returning the service object?

ThomasSchellenbergNextCentury: It's to emulate a builder pattern, so I can write "searchService.updateAggregation().updateFilter().updateSort()" etc.

rartiss55 (dismissed):

retzloff:
This change seems significant. We should be documenting major changes like this somewhere. Are we?

ThomasSchellenbergNextCentury:
@retzloff No, we're not documenting major changes right now.  I'm not too concerned about it yet because:

- Externally, we don't really have an open-source dev community.
- Internally, you're right, I need to keep our team updated, and I'm planning to do that next week with this slideshow I'm making.

I think once we're ready to publicize Neon 3 / Neon++ / (whatever) then we'll want to maintain a changelog moving forward, but I don't see it being useful until then... But maybe I'm mistaken?

retzloff:
Some things to consider for next time...

- The changes that you made will never be as fresh in your mind as there were when you made them.
- Documentation is at least, if not more, important to our team as it is to anyone outside the company. 
- Having one expert on a project - one single point of failure - is never a good idea. If something happens to make that person unavailable, then what?
- Documentation is never a priority until someone makes it one.

retzloff (approved):

PR #137, opened 2019-01-09
THOR-873

rartiss55:
Enabling visualizations to use field and table keys from dashboards.

FYI I updated the config branch a week ago but have not integrated the Angular 7 changes (holding off until it looks like those issues are resolved).

### DIFF
--- a/src/app/components/base-neon-component/base-neon.component.ts
+++ b/src/app/components/base-neon-component/base-neon.component.ts
@@ -585,40 +585,59 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
     /**
      * Updates tables, fields, and filters whenenver the database is changed and reruns the visualization query.
      *
-     * @arg {any} options A WidgetOptionCollection object.
+     * @arg {any} [options=this.options] A WidgetOptionCollection object.
      */
-    handleChangeDatabase(options: any) {
-        this.updateTablesInOptions(options);
-        this.initializeFieldsInOptions(options, this.createLayerFieldOptions());
-        this.removeAllFilters(options, this.getCloseableFilters(), () => {
+    public handleChangeDatabase(options?: any): void {
+        let optionsToUpdate = options || this.options;
+        this.updateTablesInOptions(optionsToUpdate);
+        // Change behavior depending on if the given options are the top-level options or layer options.
+        // TODO THOR-1002 How to do this nicely
+        if (optionsToUpdate === this.options) {
+            this.initializeFieldsInOptions(optionsToUpdate, this.createFieldOptions().concat(
+                new WidgetFieldOption('unsharedFilterField', 'Local Filter Field', false)
+            ));
+        } else {
+            this.initializeFieldsInOptions(optionsToUpdate, this.createLayerFieldOptions());
+        }
+        this.removeAllFilters(optionsToUpdate, this.getCloseableFilters(), () => {
             this.setupFilters();
-            this.handleChangeData(options);
+            this.handleChangeData(optionsToUpdate);
         });
     }
 
     /**
      * Updates fields and filters whenever the table is changed and reruns the visualization query.
      *
-     * @arg {any} options A WidgetOptionCollection object.
+     * @arg {any} [options=this.options] A WidgetOptionCollection object.
      */
-    handleChangeTable(options: any) {
-        this.updateFieldsInOptions(options);
-        this.initializeFieldsInOptions(options, this.createLayerFieldOptions());
-        this.removeAllFilters(options, this.getCloseableFilters(), () => {
+    public handleChangeTable(options?: any): void {
+        let optionsToUpdate = options || this.options;
+        this.updateFieldsInOptions(optionsToUpdate);
+        // Change behavior depending on if the given options are the top-level options or layer options.
+        // TODO THOR-1002 How to do this nicely
+        if (optionsToUpdate === this.options) {
+            this.initializeFieldsInOptions(optionsToUpdate, this.createFieldOptions().concat(
+                new WidgetFieldOption('unsharedFilterField', 'Local Filter Field', false)
+            ));
+        } else {
+            this.initializeFieldsInOptions(optionsToUpdate, this.createLayerFieldOptions());
+        }
+        this.removeAllFilters(optionsToUpdate, this.getCloseableFilters(), () => {
             this.setupFilters();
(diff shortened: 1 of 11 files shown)


src/app/config/config.HumanityRoads.yaml
```diff
@@ -306,7 +304,7 @@ layouts:
           bindings:
               title: Document Feed
               tableKey: table_key_1
-              idField: _id # TODO: 825: potentially make fieldKeys later
+              idField: _id
```
rartiss55, line 307: The field values are backwards compatible -- it can still accept a hardcoded value if there are no keys found. Same goes for visualization titles.

I'm a tad inconsistent about this since previously I defined tables here as 'tableKeys', so feel free to nitpick :)

src/app/dataset.ts
```diff
@@ -110,6 +110,7 @@ export class Dashboard {
     public layout?: string = '';
     public tables?: {[key: string]: string} = {};
     public fields?: {[key: string]: string} = {};
+    public visualizationTitles?: {[key: string]: string} = {};
```
rartiss55, line 113: I noticed that a couple layouts were the same in one of the configs I was using, except for a title in one of the visualizations, so I added title keys as well.

src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -1310,11 +1346,13 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
         options.table = options.tables[0] || options.table;
 
         if (options.tables.length > 0) {
-            // TODO: 873: verify which one of these is correct
-            let key = config ? config.tableKey : this.injector.get('tableKey', null);
+            let tableValue = config ? config.tableKey : this.injector.get('tableKey', null);
```
rartiss55, line 1349: I debated this for a while -- these update functions will look for a tableKey config value but if none is found, it'll use the options.table/options.database value. I also removed the array index functionality since we're moving away from arrays in our config file.

src/app/components/data-table/data-table.component.ts
```diff
@@ -299,7 +299,8 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
      */
     getColumnWidth(fieldConfig) {
         for (let miniArray of this.options.customColumnWidths) {
-            if (fieldConfig.columnName === miniArray[0]) {
+            let name = this.translateFieldKeyToValue(miniArray[0]);
```
rartiss55, line 302: Some of the non-field options in the data table use field values, so accounting for that/the presence of field keys here.

src/app/services/dataset.service.ts
```diff
@@ -222,7 +221,18 @@ export class DatasetService {
      * @return {String}
      */
     static getFieldNameByKey(dashboard: Dashboard, key: string) {
-        return dashboard.fields[key].split('.')[3];
+        let length = dashboard.fields[key].length;
```
rartiss55: Had to account for fields that contain periods.

ThomasSchellenbergNextCentury: Would it be easier just to do
```javascript
return dashboard.fields[key].split('.').slice(3).join('.');
```
I think (?) that does what you're trying to do.
(Edit:  Whoops, made a mistake the first time)

src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -584,40 +585,59 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
...
      * Updates tables, fields, and filters whenenver the database is changed and reruns the visualization query.
      *
-     * @arg {any} options A WidgetOptionCollection object.
+     * @arg {any} [options=this.options] A WidgetOptionCollection object.
      */
-    handleChangeDatabase(options: any) {
-        this.updateTablesInOptions(options);
-        this.initializeFieldsInOptions(options, this.createLayerFieldOptions());
-        this.removeAllFilters(options, this.getCloseableFilters(), () => {
+    public handleChangeDatabase(options?: any): void {
```
ThomasSchellenbergNextCentury, line 590: I think the changes to handleChangeDatabase, handleChangeTable, and handleChangeFilterField were cherry-picked from a commit to the master branch, correct?

src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -1138,7 +1158,43 @@ export abstract class BaseNeonComponent implements OnInit, OnDestroy {
...
+     * @return {any}
+     */
+    public translateFieldKeyToValue(configValue: any): string {
+        let currentDashboard = this.datasetService.getCurrentDashboard();
+
+        if (currentDashboard && currentDashboard.fields && currentDashboard.fields[configValue]) {
+            return this.datasetService.getFieldNameFromCurrentDashboardByKey(configValue);
+        } else {
+            // for backwards compatibility/if no field key reference exists in dashboard
+            return configValue;
```
ThomasSchellenbergNextCentury, line 1179: So this is going to mess up if you have a key name that's exactly the same as a field name, and the key maps to a different field, and you want to use the field, right?  I think that's an acceptable issue since it's very unlikely/stupid and we can make a note of it in the config guide.

src/app/components/base-neon-component/base-neon.component.ts
ThomasSchellenbergNextCentury: Looks good!

src/app/components/data-table/data-table.component.ts
ThomasSchellenbergNextCentury: Excellent!

src/app/config/config.HumanityRoads.yaml
ThomasSchellenbergNextCentury: That's OK.  I don't have a better suggestion for the name (besides something like "databaseAndTable" which seems too big).  Plus, "table" is already taken (and could be confusing due to the way it's currently used).

ThomasSchellenbergNextCentury (approved):

src/app/components/base-neon-component/base-neon.component.ts
rartiss55: I think I just copied a diff of the changes but yes.

src/app/services/dataset.service.ts
rartiss55: That is easier - thanks!

PR #142, opened 2019-01-21
THOR-940 THOR-946 Removed use of the filterservice REST endpoint

ThomasSchellenbergNextCentury:
Removed the use of the filterservice REST endpoint (getFilterState and messenger addFilters, removeFilters, and replaceFilters functions).  Filters are now saved in the Filter Service and are applied to all queries in the Base Neon Component.  Ensured saved state behavior still works.  This task is part of an Epic to change filter behavior within the frontend and will be merged into a feature branch as I continue to work on related tasks and cleanup (WIP).

### DIFF
--- a/src/app/services/filter.service.ts
+++ b/src/app/services/filter.service.ts
@@ -48,47 +48,6 @@ export class FilterService {
 
     constructor(protected datasetService: DatasetService) {}
 
-    protected getDatabaseFilterState(onSuccess: (filterList: any[]) => any, onError: (response: any) => any) {
-        neon.query.Filter.getFilterState('*', '*', onSuccess, onError);
-    }
-
-    /**
-     * Gets all the filters from the server.
-     * @param {Function} [onSuccess] Optional success callback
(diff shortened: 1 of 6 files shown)


src/app/services/filter.service.ts
```diff
@@ -190,22 +149,15 @@ export class FilterService {
...
-            () => {
-                for (let i = serviceFilters.length - 1; i >= 0; i--) {
-                    this.filters.push(serviceFilters[i]);
-                }
-                if (onSuccess) {
-                    onSuccess(id); // Return the ID of the primary created filter.
-                }
-            },
-            onError);
+        for (let i = serviceFilters.length - 1; i >= 0; i--) {
```
vegaed, line 152: cant this just be sandwiched in between 144-145 instead of going over the loop again.

ThomasSchellenbergNextCentury: Ah, yes, it probably can.  Thanks!

PR #144, opened 2019-01-22
THOR-720 - New typography styling

rartiss55:
Changes to font size/width throughout application to make things more consistent.


gitnordt (commented):
I noticed that there is no icon showing up in the header like it is shown in the mockup. AIDA has an icon and it would be nice to put it inline with the title of the dashboard. Is this something that we can implement in a future task? Also, the font size in the taxonomy and network graph edge labels are larger than the standard 14px.
typographyfonts_thor-720

Konnoke:
Just a note there's a difference in header style.
image

Konnoke (dismissed):
Looks good! Love the new font

rartiss55:
I added a couple updates from earlier PR comments -- please hold off on merging for now (waiting for UX approval on network graph related changes). 

**Update** - everything fine now on UX's end for network graph. Ready to be reviewed/merged.

src/app/components/aggregation/aggregation.component.scss
```diff
@@ -117,3 +117,7 @@
     }
 }
 
+.subheader {
```
ThomasSchellenbergNextCentury: The `subheader` style is repeated in multiple widgets.  Could it be moved to the `BaseNeonComponent` SCSS?

src/app/components/about-neon/about-neon.component.scss
```diff
@@ -21,4 +21,10 @@
     overflow-y: auto;
     padding-left: 10px;
     width: 100%;
+    font-size: 14px;
+}
+
+.subheader {
```
ThomasSchellenbergNextCentury: The `subheader` style here (in about-neon) is different than the `subheader` style in the widgets (the font size).  Should it be called something else to differentiate it?

ThomasSchellenbergNextCentury:
@rartiss55 In `src/app/neon-properties.scss`, `vis-title-expected-height` should be reduced from `45px` to `42px` because the `mat-toolbar` border was reduced by 3.

ThomasSchellenbergNextCentury:
@rartiss55 The change of header styles to be smaller and less-bold makes it harder for me to immediately differentiate between the header and non-header text.  It doesn't "pop" the same way.  Is it just me?  It that a silly concern?

src/app/components/aggregation/aggregation.component.scss
rartiss55: Moved this as requested.

src/app/components/about-neon/about-neon.component.scss
rartiss55: Renamed to about-subheader.

rartiss55:
@ThomasSchellenbergNextCentury in response to your earlier comments:

- Updated the vis-title-expected-height.
- I think the header text is fine since it is still bolded, but we can talk to UX about this if you'd like.

ThomasSchellenbergNextCentury (approved):

PR #149, opened 2019-02-04
Gear

Konnoke:
Added gear support for all components.


ThomasSchellenbergNextCentury:
OK, there was an issue where the diff here included commits from other (unrelated) previously merged branches, but I think that's fixed now.  Silly git.

src/app/components/gear/gear.component.ts
```diff
@@ -56,17 +70,25 @@ export class GearComponent implements OnInit, OnDestroy {
...
     ) {
         this.injector = injector;
-
+        this.collapseOptionalOptions = true;
         this.requiredList = [];
+        this.requiredListNonField = [];
         this.optionalList = [];
+        this.optionalListNonField = [];
         this.changeList = [];
+        this.mapType = 3;
```
ThomasSchellenbergNextCentury: So, we're introducing a dependency on a specific visualization (the Map) here, which is not good.  What's the issue that this solves, and what would we have to do to fix it, so we don't need to have `mapType`?

src/app/components/gear/gear.component.ts
```diff
@@ -56,17 +70,25 @@ export class GearComponent implements OnInit, OnDestroy {
     private newLimit: string;
     private changeSubcomponentType: boolean;
     private limitChanged: boolean;
+    private mapType: MapType;
+
+    public collapseOptionalOptions: boolean;
+
+    public filterVisible: Map<string, boolean> = new Map<string, boolean>();
```
ThomasSchellenbergNextCentury: I think this property determines whether the layer config options are hidden or visible?  If so, please rename to something like `hiddenLayers` because the word "filter" makes me think about a Dashboard Filter.

src/app/components/options-list/options-list.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    checkOptionType(currentType: string, checkType) {
+        if (currentType === checkType) {
+            return true;
+        }
+        return false;
+    }
+
+    ngOnInit() {
```
ThomasSchellenbergNextCentury: You can remove `ngOnInit` if you remove `implements OnInit`.  You only need to implement OnInit if you want to do something in ngOnInit but, since you don't, there's no need!

src/app/components/gear/gear.component.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+/* tslint:disable:component-class-suffix */
+
+describe('Component: Gear Component', () => {
+    let component: TestGearComponent;
+    let fixture: ComponentFixture<TestGearComponent>;
+    let getService = (type: any) => fixture.debugElement.injector.get(type);
+
+    initializeTestBed('gear component', {
+        declarations: [
+            TestGearComponent,
```
ThomasSchellenbergNextCentury: Since you don't really need the TestGearComponent for anything specific (for example, you're not overriding any functions in it), you can just use GearComponent itself and delete the TestGearComponent class.

ThomasSchellenbergNextCentury:
@Konnoke Some UX observations:
- I think I'd prefer the Cancel and Apply Changes buttons to stick on the bottom of the panel, but I'm not sure.
- The map layers should probably start expanded (right now they're collapsed by default).
- For the multi-layer options, I think the Required Config toolbar should still be shown, probably about the Add Layer button.
- I like the Required/Optional/Layer toolbars creating different sections but I think it's hard to keep track of the options in the sections sometimes.  It might be nice to see a darker background behind the options to group them (like a well:  https://www.w3schools.com/bootstrap/bootstrap_wells.asp).

ThomasSchellenbergNextCentury:
@Konnoke Oh, the Map layer options aren't populating for me (my latitude and longitude fields are showing blank when they shouldn't be).

src/app/components/options-list/options-list.component.ts
Konnoke: Thanks!

src/app/components/gear/gear.component.ts
Konnoke: HandleChangeSubcomponentType() is not consistent across different components. 

In the map component, you need pass along the new mapType while in the aggregation component, it access the new subcomponent type directly through the options.

A solution would be to change handleChangeSubcomponentType() for the map component so it uses the maptype from the option instead of passing it along as a parameter.

Konnoke: The map widget does require the new mapType in the parameter, but I realized it doesn't even use the passed mapType. In other words I wrote a lot of unnecessary code in the gear component in order to pass along the parameter when it would have been better to make the change in the map component to not require the new maptype as an argument

Konnoke:
**Changes**:

- The map layers start expanded
- For multilayer options, the required config toolbar shows above the add layer.
- There's a well for each layer so it's easier to see where the layer ends and the other required options start.
- Making changes to options inside layers now saves to the layers.
- Changed the handlechangesubcomponenttype in the map component so it's more consistent with the one in the aggregation component.
- Refactoring (ex: filterVisible to layerVisible)
- CSS update 

I left the cancel and apply button unchanged because I prefered it sticking so it's always displayed.

**Things I was unable to fix:**

The map layer options aren't populating pass the table option. The first three (title, database, and table) loads but any options after that does not even though they exists. 

I'm working on Elisa stuff now so I won't be able to work on this for a while. @rartiss55 will take over if any more changes are needed.

ThomasSchellenbergNextCentury (approved):

PR #151, opened 2019-02-07
THOR-875

rartiss55:
Fixing relations post-config file overhaul.


src/app/services/filter.service.ts
```diff
@@ -366,10 +372,12 @@ export class FilterService {
         return childFilters;
     }
 
-    protected getPermutations(originalDb: string,
+    protected getPermutations(datastore: string,
+        originalDb: string,
         originalTable: string,
-        relatedFieldMapping: Map<string, { database: string, table: string, field: string }[]>):
-        Map<{ database: string, table: string, field: string }, { database: string, table: string, field: string }>[] {
+        relatedFieldMapping: Map<string, { datastore: string, database: string, table: string, field: string }[]>):
```
rartiss55, line 378: For now I just added datastore to the relatedFieldMapping since we're adding it to the Field object later/as part of a different ticket.

ThomasSchellenbergNextCentury (approved):

PR #152, opened 2019-02-12
THOR-913 - Making sure transition partner and humanity roads configs are up to date and in new config file format

rartiss55:
For reference, examples of new format were attached to original config ticket (https://nextcentury.atlassian.net/browse/THOR-825).

### DIFF
--- a/src/app/config/config.HumanityRoads.yaml
+++ b/src/app/config/config.HumanityRoads.yaml
@@ -1,3 +1,5 @@
+# config for demo with Humanity Roads at Nepal August 15-17 2018
+
 datastores:
     es1:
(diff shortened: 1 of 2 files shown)


src/app/config/config.HumanityRoads.yaml
```diff
@@ -279,28 +268,14 @@ layouts:
           minPixely: 240
           minSizex: 3
           minSizey: 4
-        - name: Filter Builder
```
rartiss55, line 282: Removed filter builder from layout since its now in the filter tab/no longer being used as a visualization.

wagnerwj (approved):

PR #154, opened 2019-02-18
Merging LORELEI and VMAP into AIDA / VERDI (Feb 2019)

ThomasSchellenbergNextCentury:
Stacey and / or Bao, can you please do a thorough test and verify that this code works with the AIDA / VERDI dashboards / datasets ?


bdpham (dismissed):
The network graph won't filter when a node is double clicked. The filter(s) are created but the graph doesn't seem to be affected by it. I believe this bug occurred from the last neon-wide merge and not this one. I think that this should be a separate task instead of being addressed in this PR. That being said, everything else seems to work fine on my end, so I will approve. Please wait on merging/closing this PR until after @snorpet has approved.

ThomasSchellenbergNextCentury:
@bdpham Concerning the filter bug:  This happens on verdi-master too.  I think the behavior was added intentionally.  Here's the relevant line:  https://github.com/NextCenturyCorporation/neon-dash-internal/blame/verdi-master/src/app/components/network-graph/network-graph.component.ts#L712

eaciv:
@bdpham, @ThomasSchellenbergNextCentury: Oof...this was caused by me? I introduced that condition to avoid entering code I didn't think it would need to with an empty result set. My code will work correctly if the "else if" condition is removed. Sorry for the inconvenience, guys.

src/app/components/query-bar/query-bar.component.ts
```diff
@@ -112,18 +112,19 @@ export class QueryBarComponent  extends BaseNeonComponent {
...
      */
-    finalizeVisualizationQuery(options: any, query: neon.query.Query, wherePredicates: neon.query.WherePredicate[]): neon.query.Query {
-        let wheres: neon.query.WherePredicate[] = wherePredicates.concat(neon.query.where(options.filterField.columnName, '!=', null));
-        return query.where(wheres.length > 1 ? neon.query.and.apply(neon.query, wheres) : wheres[0])
-            .sortBy(options.filterField.columnName, neonVariables.ASCENDING);
+    finalizeVisualizationQuery(options: any, query: NeonQueryPayload, sharedFilters: NeonFilterClause[]): NeonQueryPayload {
+        let filter: NeonFilterClause = this.searchService.buildFilterClause(options.filterField.columnName, '!=', null);
+        this.searchService.updateFilter(query, this.searchService.buildBoolFilterClause(sharedFilters.concat(filter)))
+            .updateSort(query, options.filterField.columnName);
+        return query;
```
gitnordt, line 127: Query Bar is no longer cleared out when all filters are removed. I'm not sure if this code caused the issue or if it was a prior merge. We may need to put a separate ticket in for this one.

src/app/components/media-viewer/media-viewer.component.html
```diff
@@ -157,32 +158,36 @@
...
+                             [class]="tab.selected.border ? ('image with-border border-mat-' + tab.selected.border + ' medium') : 'image medium'">
+                        <img *ngIf="tab.selected.mask" [src]="tab.selected.mask" [alt]="tab.selected.name" id="mask0" [hidden]="!tab.loaded"
+                             [class]="tab.selected.border ? ('image with-border border-mat-' + tab.selected.border + ' mask') : 'image mask'">
+                        <mat-slider color="primary" min="0" max="100" step="1" thumbLabel [value]="options.sliderValue"
+                                    (input)="onSliderChange($event.value)"></mat-slider>
+                    </div>
+                    <img *ngSwitchCase="mediaTypes.image" [src]="tab.selected.link" [alt]="tab.selected.name"
+                        [class]="tab.selected.border ? ('image with-border border-mat-' + tab.selected.border) : 'image'">
+                    <video *ngSwitchCase="mediaTypes.video" controls
+                        [class]="tab.selected.border ? ('video with-border border-mat-' + tab.selected.border) : 'video'">
```
gitnordt: AIDA has a recurring issue with youtube videos not showing up in the media viewer (referencing AIDA-235). The fix for this is to remove the video html tag as it can not handle youtube embed links. The iframe tag should be used to handle all videos instead. Unfortunately, even though I have removed the video tags and related code from the verdi-master repo multiple times, the code keeps getting merged back in causing the youtube links to break again. Please remove the video tags and related code.

src/app/components/current-filters/current-filters.component.html
```diff
@@ -1,25 +1,25 @@
...
+        <mat-icon>delete_sweep</mat-icon>
+    </button>
+</div>
+
+<div>
+    <div mat-fill>
+        <mat-list>
+            <mat-list-item *ngFor="let filter of filters.formatted" class="current-filters-item filter-text">
+                <button mat-icon-button matTooltip="Remove filter" matTooltipPosition="below" (click)="removeFilter(filter.ids)">
+                    <mat-icon>delete</mat-icon>
```
gitnordt, line 13: I found an issue that is a little difficult to recreate, but I am getting some inconsistent behavior between the filters for the taxonomy and network graph. Sometimes deleting all filters does not reset the components.

One case where I seem to be able to recreate the issue: Go to the TA2 Dashboard. Click on a document to filter network graph, then double click on a node in the network graph to add additional filters. Then, remove the filters from the network graph(Not from the filter tray). The network graph is still filtered even though all filters have been removed.

I wasn't sure if this needed to be a separate bug or if it could be addressed in these changes.

gitnordt (changes requested):

src/app/components/media-viewer/media-viewer.component.html
ThomasSchellenbergNextCentury: @snorpet VMAP needs to be able to show videos.  Let's get together to talk about a collective solution.

src/app/components/map/map.component.html
```diff
@@ -44,7 +44,7 @@
                 <p></p>
 
                 <div *ngFor="let layer of options.layers">
-                    <mat-toolbar style="cursor: pointer;" (click)="toggleFilter(layer)">
+                    <mat-toolbar class="subheader" style="cursor: pointer;" (click)="toggleFilter(layer)">
                         <span>{{ layer.title }}</span>
                         <span class="toolbar-spacer"></span>
                         <mat-icon class="neon-icon-small" *ngIf="options.layers.length > 1" (click)="removeLayer(layer)">delete</mat-icon>
```
gitnordt, line 50: I am not sure if this is an AIDA-specific issue or if other projects are seeing this as well. I noticed that when clicking on the layers icon on the map, the menu contains a checkbox input with a label that says 'undefined'. 

I wasn't sure if this was a product of the merge or something that has been there.

gitnordt (changes requested):

src/app/components/current-filters/current-filters.component.html
ThomasSchellenbergNextCentury: I'm having trouble reproducing the issue, but it may be related to the bug that Bao found.  Let me implement the fix that Eddie recommended and you can try it again.

src/app/components/query-bar/query-bar.component.ts
ThomasSchellenbergNextCentury: It was definitely caused by a prior merge (I tested verdi-master).  I don't know of a quick-fix, so let's add a ticket for this in the backlog.  On a related note, maybe I'll be able to fix this in the filter changes I've been doing, but I don't want to guarantee anything right now :)

src/app/components/map/map.component.html
ThomasSchellenbergNextCentury: As we discussed earlier, let's worry about this later.  Who knows how long it's been there!

gitnordt: Put in a ticket for this issue so we can keep track, NEON-12

src/app/components/query-bar/query-bar.component.ts
gitnordt: Put in a ticket for this one, AIDA-560

gitnordt (dismissed):

bdpham (approved):
My requested change has been fixed. Approving.

PR #164, opened 2019-03-05
THOR-1039

Konnoke:
The filter builder now shows the database of the current dashboard whereas before it was always the first database.

New widget defaults to the current database of the current dashboard whereas before it was always the first database.

### DIFF
--- a/src/app/services/dataset.service.ts
+++ b/src/app/services/dataset.service.ts
@@ -527,6 +527,28 @@ export class DatasetService {
     }
 
     /**
+     * Returns the database with the given Dashboard name or an Object with an empty name if no such database exists in the dataset.
+     * @param {String} The dashboard name
+     * @return {Object} The database containing {String} name, {Array} fields, and {Object} mappings if a match exists
+     * or undefined otherwise.
+     * Dashboard name only includes part of the database pretty name
+     */
+    public getDatabaseWithDashboardName(): DatabaseMetaData {
+        if (!this.getCurrentDashboard()) {
+            return undefined;
+        }
+        let tableKeys = this.getCurrentDashboard().tables;
+
+        let keyArray = Object.keys(tableKeys);
+
+        if (keyArray.length) {
+            let databaseName = this.getDatabaseNameFromCurrentDashboardByKey(keyArray[0]);
+            return this.getDatabaseWithName(databaseName);
+        }
+        return undefined;
+    }
+
+    /**
      * Returns the tables for the database with the given name in the active dataset.
      * @param {String} The database name
      * @return {Array} An array of table Objects containing {String} name, {Array} fields, and {Array} mappings.
--- a/src/app/services/dataset.service.ts
+++ b/src/app/services/dataset.service.ts
@@ -533,7 +533,7 @@ export class DatasetService {
      * or undefined otherwise.
      * Dashboard name only includes part of the database pretty name
      */
-    public getDatabaseWithDashboardName(): DatabaseMetaData {
+    public getCurrentDatabase(): DatabaseMetaData {
         if (!this.getCurrentDashboard()) {
             return undefined;
         }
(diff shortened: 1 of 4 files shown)


rartiss55:
So the case where adding a new visualization and seeing a database for the current dashboard works, but there are still issues with the filter tray -- the only time I've consistently seen it update correctly is when I wait to open it for the first time until after I've updated my dashboard. 

If you:
1.) Pick dashboard 1
2.) Open and close filter tray
3.) Pick dashboard 2

The database in the filter tray isn't updated. Additionally if I selected filters on a previous dashboard and picked a new one, they still appear in the Active Filters section of the filter tray.

src/app/services/dataset.service.ts
```diff
@@ -526,6 +526,28 @@ export class DatasetService {
...
     }
 
+    /**
+     * Returns the database with the given Dashboard name or an Object with an empty name if no such database exists in the dataset.
+     * @param {String} The dashboard name
+     * @return {Object} The database containing {String} name, {Array} fields, and {Object} mappings if a match exists
+     * or undefined otherwise.
+     * Dashboard name only includes part of the database pretty name
+     */
+    public getDatabaseWithDashboardName(): DatabaseMetaData {
```
rartiss55: Since this is no longer matching the database with part of the dashboard pretty name, should the name/docs be updated here? Maybe something like "getDatabaseForCurrentDashboard"?

ThomasSchellenbergNextCentury:
Some ideas:
- The `messenger` in the `FilterBuilder` can listen for the `DASHBOARD_CLEAR` event and remove all elements from its `clauses` array.
- After the dashboard is updated, a call to the FilterBuilder's `addBlankFilterClause` should initialize it properly.

Konnoke:
Currently doing the following, 
1.) Pick dashboard 1
2.) Open and close filter tray
3.) Pick dashboard 2
4.) Open filter tray

will result in showing the correct fields in the filter builder but database and table will be wrong. I tried testing DASHBOARD_CLEAR channel and the NEW_DATASET messenger channel to call the new clearFilterBuilder function I made but the updateDatabaseInOptions in base neon and getCurrentDatabase in the datasetService still seems to return the old/first database.

ThomasSchellenbergNextCentury:
@Konnoke I believe I fixed the issue; please confirm.

Here's what I did:
- With Chrome's dev tools, I traced the behavior of the Filter Builder when the dashboard changed, starting at clearFilterBuilder.  I saw that the filter clauses were removed and a new blank filter clause was added as expected.  However, the database and table of the blank filter clause were being changed in addBlankFilterClause using the Filter Builder's internal options; since the Filter Builder is no longer its own widget, its internal options aren't accurate, so I removed that part.
- With that fix, the clauses array was being changed correctly, but the dropdown element wasn't being updated correctly.  I noticed that, if I interacted with the Filter Builder (by clicking an input field or button), the dropdown would immediately update; this is an indication that Angular had not been notified of data model changes.  I added a changeDetection.detectChanges to clearFilterBuilder and that solved the issue.

Please let me know if you have any questions.

Konnoke:
@ThomasSchellenbergNextCentury What do you mean by the Filter Builder isn't its own widget? I know that the filter builder was removed from add visualization but it still extends base neon. What changed so that the internal options aren't accurate

rartiss55 (approved):
@ThomasSchellenbergNextCentury looks like its working now. Approving.

ThomasSchellenbergNextCentury:
@Konnoke Sorry, I wasn't clear.  I use "widget" as a term for the components within the grid layout.  The Filter Builder is no longer defined as a component within the grid layout; now, it's a permanent fixture to all dashboards through the navbar.  Additionally, the Filter Builder is no longer configured using the config file; since the `options` property is used to configure widgets from the config file, it doesn't really make sense any more for the Filter Builder, so we can ignore it in this case.  (In fact, we should remove Filter Builder as a subclass of BaseNeonComponent sometime, but that's a big task for another day.)

ThomasSchellenbergNextCentury (approved):

PR #165, opened 2019-03-06
THOR-1038: Filtering the map using the legend should be working again.

rartiss55:
Previously showing/hiding points using the legend on the map was throwing a console error. Should be fixed now.


Konnoke (approved):
Ditto

ThomasSchellenbergNextCentury (approved):

PR #168, opened 2019-03-11
THOR-1046 Fix Database, Table, and Layer Changes in Gear Component

ThomasSchellenbergNextCentury:
Currently, changing the database or table in the widget option (gear) menu immediately updates the widget's options; it does not wait for you to click Apply and is not affected by Cancel.  The same also happened with adding, removing, and changing options of Layers in the Map.

1. I changed the Gear component to create a second WidgetOptionCollection to record modifications (adding `modifiedOptions`) rather than a list of "change" objects (removing `changeList` and `changeLayerList`).

2. I moved the `updateDatabases`, `updateTables`, `updateFields`, and `findField` functions from the BaseNeonComponent to the WidgetOptionCollection (with corresponding unit tests).  The Gear component now calls these functions on the WidgetOptionCollection rather than passing the functions in the event message.  (I also moved `translateFieldKeyToValue` from the BaseNeonComponent to the DatasetService).

3. I separated the BaseNeonComponent's `addLayer` and `removeLayer` functions into multiple functions (one to initialize the change, a second to update the widget itself) that are sent individually to the Gear component.

4. I added static dropdowns for the title, database, and table for layers in the Gear component, changed input element "name" attributes to be unique, and removed some unused functions.

5. I changed the OptionList to accept a WidgetOptionCollection, list of binding keys (to show from the WidgetOptionCollection), and list of fields (needed because, if I just used `options.fields`, the the dropdowns wouldn't update).  I also removed the database and table dropdowns (since they're in the Gear component now).

6. I updated the DatasetServiceMock to create real classes to work with the new DatasetService unit tests.

### DIFF
--- a/src/app/components/annotation-viewer/annotation-viewer.component.ts
+++ b/src/app/components/annotation-viewer/annotation-viewer.component.ts
@@ -876,7 +876,7 @@ export class AnnotationViewerComponent extends BaseNeonComponent implements OnIn
 
             // This will ignore a filter with multiple clauses.
             if (!neonFilter.filter.whereClause.whereClauses) {
-                let field = this.findField(this.options.fields, neonFilter.filter.whereClause.lhs);
+                let field = this.options.findField(neonFilter.filter.whereClause.lhs);
                 let value = neonFilter.filter.whereClause.rhs;
                 if (this.isVisualizationFilterUnique(field.columnName, value)) {
                     this.addVisualizationFilter(this.createVisualizationFilter(neonFilter.id, field.columnName, field.prettyName, value));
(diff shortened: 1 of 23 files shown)


Konnoke (approved):

Konnoke:
There doesn't seem to be any "none" options and optional fields seem to have an * next to them

rartiss55:
What @Konnoke said -- other than that looks good!

ThomasSchellenbergNextCentury:
Fixed; see https://github.com/NextCenturyCorporation/neon-dash-internal/commit/eefc7b95f4ccd435f7cd4187246d688640340d31

PR #172, opened 2019-03-15
THOR-1074 Removed isMultiLayerWidget property because it is redundant.

ThomasSchellenbergNextCentury:
Removed `isMultiLayerWidget` from the BaseNeonComponent on Eddie's suggestion.  Most of the time it is redundant with checking if `options.layers.length > 0`.  Users can no longer remove the final layer of a component with existing layers.  I also added the `shouldCreateDefaultLayer` function.


Konnoke (approved):

eaciv:
I can't really speak to most of the changes, but the removal of isMultiLayerWidget looks reasonable. Luckily, it wasn't used to too many components ;)

PR #178, opened 2019-04-02
THOR-1025

rartiss55:
Fix for incorrect legend color in hover element for aggregation charts (was displaying just a black square instead of the actual color shown on the chart).

### DIFF
--- a/src/app/color.ts
+++ b/src/app/color.ts
@@ -154,10 +154,10 @@ export class Color {
         if (colorCss.indexOf('var(--') === 0) {
             let css = colorCss.substring(4, colorCss.length - 1);
             css = css.indexOf(',') >= 0 ? css.substring(0, css.indexOf(',')) : css;
-            return getComputedStyle(elementRef.nativeElement).getPropertyValue(css);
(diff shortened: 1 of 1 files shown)


src/app/color.ts
```diff
@@ -154,10 +154,10 @@ export class Color {
         if (colorCss.indexOf('var(--') === 0) {
             let css = colorCss.substring(4, colorCss.length - 1);
             css = css.indexOf(',') >= 0 ? css.substring(0, css.indexOf(',')) : css;
-            return getComputedStyle(elementRef.nativeElement).getPropertyValue(css);
+            return getComputedStyle(elementRef.nativeElement).getPropertyValue(css).trim();
```
rartiss55, line 157: Used trim() instead of a global replace since the issue was only with leading/trailing spaces.

Konnoke (approved):
Huge

ThomasSchellenbergNextCentury (approved):

ThomasSchellenbergNextCentury:
Such code.  Much excite.  Wow.

PR #180, opened 2019-04-04
Thor 501

Konnoke:
**THOR-501:** 
Add "no data" message to charts and page.
This is for when user has filtered chart to point where there is no data to display.

**What's new:**
- Data Message Component that shows an icon and text
- Added new functions and fields to baseNeonComponent to detect when there's no active data after filtering. 
- Added a new function toggleBodyContainer in baseNeonComponent that is override in the aggregation component so it can hide the body container when it's showing the Data Message Component.
- Added Data Message Component to Aggregation, Media Viewer, and Textcloud.
- Updated unit tests

**Attempted but not included because of issues:**
Tried adding more comprehensive tests by making a TestBaseNeonComponentMock class that extends TestBaseNeonComponent but uses the mock services in order to test the behaviors when filters are added or removed.

image

### DIFF
--- a/src/app/components/aggregation/aggregation.component.ts
+++ b/src/app/components/aggregation/aggregation.component.ts
@@ -1329,7 +1329,6 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
      * @override
      */
     subcomponentRequestsRedraw(event?) {
-        this.noDataCheck();
         if (event) {
             this.stopEventPropagation(event);
         }
--- a/src/app/components/aggregation/aggregation.component.ts
+++ b/src/app/components/aggregation/aggregation.component.ts
@@ -1368,12 +1368,7 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
      */
     toggleBodyContainer() {
         let bodyContainer: HTMLElement = document.getElementById('body-container');
-
-        if (this.showNoData) {
-            bodyContainer.setAttribute('style', 'display: none');
-        } else {
-            bodyContainer.setAttribute('style', 'display: show');
-        }
+        bodyContainer.setAttribute('style', 'display: ' + (this.showNoData ? 'none' : 'show'));
     }
 
     /**
--- a/src/app/components/aggregation/aggregation.component.ts
+++ b/src/app/components/aggregation/aggregation.component.ts
@@ -1363,6 +1363,20 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
     }
 
     /**
+     * Toggles the body container if the data message component is present
+     * @override
+     */
+    toggleBodyContainer() {
+        let bodyContainer: HTMLElement = document.getElementById('body-container');
+
+        if (this.showNoData) {
+            bodyContainer.setAttribute('style', 'display: none');
+        } else {
+            bodyContainer.setAttribute('style', 'display: show');
+        }
+    }
+
+    /**
      * Toggles the given filter in the given filter list and recreates or removes the neon filter.
      *
      * @arg {Filter} filter
--- a/src/app/components/aggregation/aggregation.component.ts
+++ b/src/app/components/aggregation/aggregation.component.ts
@@ -1450,4 +1450,9 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
             (options.aggregation !== AggregationType.COUNT ? options.aggregationField.columnName : true);
         return !!(options.database.name && options.table.name && validFields);
     }
+
+    protected clearVisualizationData(options: any): void {
+        // TODO THOR-985 Temporary function.
+        this.onChangeData();
+    }
 }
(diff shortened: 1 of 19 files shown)


src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -1403,4 +1405,30 @@ export abstract class BaseNeonComponent implements AfterViewInit, OnInit, OnDest
...
+     * Checks wheather there are any filters and returns no data.
+     */
+    public noDataCheck() {
+        if (this.filterService.getFilters().length > 0) {
+            let activeData = this.getActiveData();
+            if (!activeData || !activeData.data || !activeData.data.length) {
+                this.showNoData = true;
+            } else {
+                this.showNoData = false;
+            }
```
ThomasSchellenbergNextCentury: FYI You don't have to change this, but this could be handled in fewer lines:
```
this.showNoData = !activeData || !activeData.data || !activeData.data.length;
```

src/app/components/aggregation/aggregation.component.ts
```diff
@@ -1362,6 +1362,20 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
...
+     * @override
+     */
+    toggleBodyContainer() {
+        let bodyContainer: HTMLElement = document.getElementById('body-container');
+
+        if (this.showNoData) {
+            bodyContainer.setAttribute('style', 'display: none');
+        } else {
+            bodyContainer.setAttribute('style', 'display: show');
+        }
```
ThomasSchellenbergNextCentury: FYI You don't have to change this, but this could be handled in fewer lines:
```
bodyContainer.setAttribute('style', 'display: ' + (this.showNoData ? 'none' : 'show'));
```

src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -1403,4 +1405,30 @@ export abstract class BaseNeonComponent implements AfterViewInit, OnInit, OnDest
...
+        this.changeDetection.detectChanges();
+        this.toggleBodyContainer();
+    }
+
+    /**
+     * Method to be overrided in components where the body container wants to be hidden
+     * if showNoData is true
+     */
+    public toggleBodyContainer() {
+        //Will be override in the aggregation component
```
ThomasSchellenbergNextCentury: You shouldn't name specific subclasses like this in override-able functions in a superclass:  you never know when you will create a new subclass that needs to override the same function, and then the comment will be stale.

src/app/components/data-message/data-message.component.ts
```diff
@@ -0,0 +1,60 @@
...
+    selector: 'app-data-message',
+    templateUrl: 'data-message.component.html',
+    styleUrls: ['data-message.component.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    encapsulation: ViewEncapsulation.Emulated
+})
+export class DataMessageComponent implements OnInit {
+    public chartsAndGraph: any[];
+    public visualizations: any[];
+    public currentVisualization: any;
```
ThomasSchellenbergNextCentury: It doesn't look like these three variables (`chartsAndGraph`, `visualizations`, and `currentVisualization`) are used anywhere.  If that's really the case, please remove them.

ThomasSchellenbergNextCentury:
It seems like the "No Data" message would be useful in other visualizations like the Document Viewer, Network Graph, News Feed, or Thumbnail Grid.  Should we add it to them too, or is that outside the scope of this task?

ThomasSchellenbergNextCentury:
I fixed an issue I found (an existing bug, unrelated to anything you did!) so the aggregation component would clear its legend properly.

ThomasSchellenbergNextCentury (changes requested):
Please review my previous comments.

Konnoke:
> It seems like the "No Data" message would be useful in other visualizations like the Document Viewer, Network Graph, News Feed, or Thumbnail Grid. Should we add it to them too, or is that outside the scope of this task?

Yep, that's currently outside the scope of the ticket. The ticket only mentioned "charts and page" which Sophie clarified to mean only the aggregation, media viewer, and text cloud.

ThomasSchellenbergNextCentury:
OK, cool.  I created a new ticket to cover the other visualizations:  https://nextcentury.atlassian.net/browse/THOR-1117

src/app/components/data-table/data-table.component.spec.ts
```diff
@@ -23,6 +23,7 @@ import { NgxDatatableModule } from '@swimlane/ngx-datatable';
 import { } from 'jasmine-core';
 
 import { NeonGTDConfig } from '../../neon-gtd-config';
+import { DataMessageComponent } from '../data-message/data-message.component';
```
ThomasSchellenbergNextCentury: You don't need the import here since it's not used in the Data Table.

Konnoke: Nice catch! I forgot to remove that.
I initially had it in the data table but I removed it because it was redundant. Ngx-datatable had its own message for no data.

ThomasSchellenbergNextCentury (approved):

PR #182, opened 2019-04-09
THOR-1013 - Fix issue in data table where fields weren't updating when database/table is changed.

rartiss55:


ThomasSchellenbergNextCentury:
@rartiss55 So I can see that, when I change the database in the data table, the column selector list does change.  However, I would also expect the headers in the data table itself to change, but they don't, and that seems like a bug to me.  Can you see that too?  Is that something that is outside the scope of this ticket, or should we fix it now?

src/app/components/data-table/data-table.component.ts
```diff
@@ -186,7 +186,8 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
             }]),
             new WidgetNonPrimitiveOption('customColumnWidths', 'Custom Column Widths', [], false),
             new WidgetNonPrimitiveOption('exceptionsToStatus', 'Exceptions to Status', [], false),
-            new WidgetNonPrimitiveOption('fieldsConfig', 'Fields Config', {})
+            new WidgetNonPrimitiveOption('fieldsConfig', 'Fields Config', {}),
+            new WidgetSelectOption('databaseOrTableChange', 'Database or Table Updated', false, OptionChoices.NoFalseYesTrue, false)
```
ThomasSchellenbergNextCentury: I'm concerned about the way we're signaling the database/table change here.  Adding the new `databaseOrTableChange` variable introduces a few complications that make me uncomfortable:
1. We're adding a visualization-specific dependency to the GearComponent (on a variable that only exists within the Data Table).
2. We're adding a config option that we don't intend to use in the config file.
3. We're saving a temporary status in a permanent variable.

If we want to consider alternative solutions, here are a few suggestions:
1. Add a "databaseOrTableChange" argument to the `handleChangeData`, `handleChangeFilterField`, and `onChangeData` functions, and have the GearComponent pass the databaseOrTableChange status to the Data Table using the existing callback functions.
2. Like # 1, but, instead of a "databaseOrTableChange" boolean, have a list of option keys that were changed.  The Data Table would know of a database/table change if "database" or "table" were in the list.  This would be a bit more complex but could be more versatile in the future.  (This may even let us remove handleChangeFilterField since the visualization itself can determine whether a filter field was changed.)
3. Add a new callback to the "options" event published by the BaseNeonComponent that the GearComponent can call when the database/table is changed.  I would prefer 1 or 2 over this option because I don't want to bloat the event message with a lot of callbacks.

What do you think?  Am I missing something?  We can discuss the situation in person tomorrow if you'd prefer.  Thanks!

ThomasSchellenbergNextCentury (changes requested):
See comments.

rartiss55:
> @rartiss55 So I can see that, when I change the database in the data table, the column selector list does change. However, I would also expect the headers in the data table itself to change, but they don't, and that seems like a bug to me. Can you see that too? Is that something that is outside the scope of this ticket, or should we fix it now?

@ThomasSchellenbergNextCentury Ah - so for me, when I change the database, the sortField unpopulates. I have to select a new field in that list since its required to get the data table to requery. By doing that, the headers should update. If we don't like that behavior, I think we should log a separate ticket.

src/app/components/data-table/data-table.component.ts
rartiss55: Redid it as option 1.

ThomasSchellenbergNextCentury (approved):
Looks better.  Thanks for making the requested changes.

PR #183, opened 2019-04-09
Aida 555

gitnordt:
I added leaf nodes at the lowest level of the taxonomy to represent the name values in the AIDA seedling data. The leaf nodes do not have checkboxes or filtering capability like the other nodes in the taxonomy. They simply exist as a reference for the user. The leaf node level is styled as a scrolling div for separation, neatness, and legibility. The values can be hidden or shown using the settings menu in the taxonomy widget.

### DIFF
--- a/src/app/components/taxonomy-viewer/taxonomy-viewer.component.html
+++ b/src/app/components/taxonomy-viewer/taxonomy-viewer.component.html
@@ -60,7 +60,7 @@
                 <p></p>
 
                 <mat-form-field>
-                    <mat-select placeholder="Parent Field" [(ngModel)]="options.categoryField" required="true"
+                    <mat-select placeholder="Root Field" [(ngModel)]="options.categoryField" required="true"
                                 (ngModelChange)="handleChangeData()" [disabled]="options.fields.length < 1">
                         <mat-option [value]="createEmptyField()">(None)</mat-option>
                         <mat-option *ngFor="let field of options.fields" [value]="field">{{ field.prettyName }}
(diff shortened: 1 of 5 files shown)


ThomasSchellenbergNextCentury:
This looks good!  I'm assuming the leaf nodes don't filter because the number of filters could cause performance issues?

ThomasSchellenbergNextCentury:
@snorpet FYI The leaf nodes don't show up if your typeField is not set.  I'm not sure if that's intentional or not.  Regardless, I'm assuming you can wait to fix it if/when it becomes relevant :)

ThomasSchellenbergNextCentury (dismissed):

gitnordt:
> This looks good! I'm assuming the leaf nodes don't filter because the number of filters could cause performance issues?

This was a design decision made by @s0sterUX and I, but it definitely makes sense to limit the filtering capability for the timeline. Once all of the outstanding filtering issues are resolved, we may want to revisit this concept.

gitnordt:
> @snorpet FYI The leaf nodes don't show up if your typeField is not set. I'm not sure if that's intentional or not. Regardless, I'm assuming you can wait to fix it if/when it becomes relevant :)

This is a good observation. I am inclined to say that this is intentional. If you only have a categoryField then one node level would be displayed. Technically, the typeField could be used to display the leaf node values if you wanted values listed under the categoryFields. This level of values would be filterable. Then there is a subTypeField that could potentially be a third level of filterable values. Therefore, it's safe to say that the leaf node level is a tertiary or quaternary level for values that you intentionally do not want to filter. If there is a circumstance in which leaf nodes need to be displayed directly under the categoryField then we can make the code change, but I don't foresee this being an issue.

Thank you for raising this point. In you bringing this to my attention, I made a few modifications to the code to eliminate a few assumptions that I made previously.

ThomasSchellenbergNextCentury (approved):

PR #184, opened 2019-04-15
THOR-1115: fix for layers being labeled as 'undefined' in overlay control.

rartiss55:
Layer title should now match what's specified in the gear component (see original ticket for screenshot of the issue).

### DIFF
--- a/src/app/components/map/map.type.leaflet.ts
+++ b/src/app/components/map/map.type.leaflet.ts
@@ -83,8 +83,16 @@ export class LeafletNeonMap extends AbstractMap {
         }
     }
 
-    addPoints(points: MapPoint[], layer?: any, cluster?: boolean) {
+    addPoints(points: MapPoint[], layer?: any, cluster?: boolean, layerTitle?: string) {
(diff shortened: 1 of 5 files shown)


src/app/components/map/map.type.leaflet.ts
```diff
@@ -104,7 +112,7 @@ export class LeafletNeonMap extends AbstractMap {
                 }
             });
             this.layerGroups.set(layer, layerGroup);
-            this.layerControl.addOverlay(layerGroup, layer.title);
+            this.layerControl.addOverlay(layerGroup, layerTitle);
```
rartiss55, line 115: Originally title property was referenced as something that existed as part of the layer. I initially added it as part of the layer property (and updated all keys to be something like {id: '123', title: 'Layer 1'}, but took it out and left it as before (with just the id being used as the map layer/layer key) since we can update the title anytime in the gear component, which made the combined id/title approach a bit more complicated.

ThomasSchellenbergNextCentury (approved):
Looks awesome!

PR #185, opened 2019-04-18
Aida 627

gitnordt:
Implemented the ability to add and filter by an additional field in the timeline. This allows for cross-component filtering on multiple fields. These changes are dependent upon the most recent changes to the Neon Server code ([NextCenturyCorporation/neon-server](https://portal.nextcentury.com/owncloud/index.php/s/lRpjvIBXmfszWuz)). You will need to deploy the new Neon Server war file, before testing this PR.


bdpham (changes requested):
Tests are failing:
ReferenceError: d3 is not defined

gitnordt:
> Tests are failing:
> ReferenceError: d3 is not defined

Added d3 import back in, it got removed when I was cleaning up my code. Sorry.

src/app/components/timeline/timeline.component.ts
```diff
@@ -316,19 +397,114 @@ export class TimelineComponent extends BaseNeonComponent implements OnInit, OnDe
...
+            data = results.reduce((itemBucket, currentItem) => {
+                let previousItem = this.findDateInPreviousItem(itemBucket, currentItem);
+                if (previousItem) {
+                    if (previousItem.ids.indexOf(currentItem[options.idField.columnName]) === -1) {
+                        previousItem.ids.push(currentItem[options.idField.columnName]);
+                        previousItem.value++;
+                    }
+
+                    previousItem.filters = previousItem.filters.concat(currentItem[options.filterField.columnName])
+                        .filter((value, index, array) => array.indexOf(value) === index);
```
ThomasSchellenbergNextCentury, line 420: This `filter` is to remove duplicate values in `previousItem.filters`, right?

src/app/components/timeline/TimelineSelectorChart.ts
```diff
@@ -875,7 +879,13 @@ export class TimelineSelectorChart {
                 _.debounce(() => {
                     // Update the chart
                     this.redrawChart();
-                    this.tlComponent.onTimelineSelection(this.data.extent[0], this.data.extent[1]);
+                    for (let data of this.data.data[0].data) {
+                        if (data.date >= this.data.extent[0] && data.date <= this.data.extent[1]) {
```
ThomasSchellenbergNextCentury: I think there's a bug here:  the `<=` should be a `<`.

As we've discussed before, the timeline's filter is inclusive on the start date but exclusive on the end date (`>= startDate` and `< endDate`).

I noticed that, whenever you filter, the ID filters (or whatever is marked as the `filterField`) are including data that belongs to an extra bar in the timeline.  For example, if you filter the date on a single bar, it filter the IDs on both that bar AND the bar directly to its right.  Tracing through the code, it looks like the `onTimelineSelection` function was being sent too much `selectedData`, and changing the `<=` to a `<` seemed to fix the issue.

Let me know if my understanding is mistaken or if this solution has unexpected side-effects.

src/app/components/timeline/timeline.component.ts
gitnordt: That is correct.

src/app/components/timeline/TimelineSelectorChart.ts
gitnordt: Thank you for catching this. I change it in one part of the code, but did not make the change here.

ThomasSchellenbergNextCentury (dismissed):

bdpham (commented):
There are a few filtering bugs that I notice. @snorpet and I have discussed this and will be working to get them fixed.

gitnordt:
> There are a few filtering bugs that I notice. @snorpet and I have discussed this and will be working to get them fixed.

I had a difficult time recreating the bug that Bao found. I had to change the granularity to day first, but it looks like if you select a part of the timeline that does not have a bar and then select a part of the timeline that does have a bar, the added filter field value no longer updates. I was able to confirm this in the active filters. I need to figure out how to capture the values for the filter field after an empty selection is made. I will work on the solution and update the code.

gitnordt:
> > There are a few filtering bugs that I notice. @snorpet and I have discussed this and will be working to get them fixed.
> 
> I had a difficult time recreating the bug that Bao found. I had to change the granularity to day first, but it looks like if you select a part of the timeline that does not have a bar and then select a part of the timeline that does have a bar, the added filter field value no longer updates. I was able to confirm this in the active filters. I need to figure out how to capture the values for the filter field after an empty selection is made. I will work on the solution and update the code.

The issue was that once there was a selection on the timeline that returned no data, there was a filter created without a whereClause. This posed a problem when trying to locate/replace the filter since it did not include a whereClause associated with the filterField. After speaking with Thomas, we found that the best way to approach this issue is to create a filter with a whereClause that sets the filterField equal to an empty string. That way it can be located by the whereClause (lhs) and replaced as needed. 

Filtering should work as expected now. When a selection on the timeline returns no data, all other components associated with that filter field should be cleared.

bdpham (dismissed):

src/app/components/timeline/timeline.component.ts
```diff
@@ -214,19 +214,18 @@ export class TimelineComponent extends BaseNeonComponent implements OnInit, OnDe
...
-            };
+                let setFilter = {
+                    id: undefined,
+                    field: this.options.filterField.columnName,
+                    prettyField: this.options.filterField.prettyName,
+                    value: selectedFilters
+                };
 
-            this.manageFieldFilters(setFilter, neonFilters, false);
+                this.manageFieldFilters(setFilter, neonFilters, false);
```
ThomasSchellenbergNextCentury: The indentation on lines 221-228 looks like it got messed up.

src/app/components/timeline/timeline.component.ts
```diff
@@ -255,8 +254,12 @@ export class TimelineComponent extends BaseNeonComponent implements OnInit, OnDe
     createNeonFieldFilter(filterField: string, filterValues: string[]): neon.query.WherePredicate {
         let clauses = [];
 
-        for (let value of filterValues) {
-            clauses.push(neon.query.where(filterField, '=', value));
+        if (!filterValues.length) {
+            clauses.push(neon.query.where(filterField, '=', ''));
```
ThomasSchellenbergNextCentury, line 259: Could you please add a comment here that the "filter on field equals empty string" behavior might not work as expected with every dataset (basically what we discussed yesterday)?  You can mention ticket NEON-36 too.

src/app/components/timeline/timeline.component.ts
gitnordt: Done

src/app/components/timeline/timeline.component.ts
gitnordt: Done

bdpham (approved):

ThomasSchellenbergNextCentury (approved):

PR #186, opened 2019-04-19
THOR-942 Filter Refactor (Aggregation Component)

ThomasSchellenbergNextCentury:

HOW DOES THOR-942 FIX IT

WHAT IS IN THIS PR
- Updates to the Aggregation Component to make it work with the new FilterService and BaseNeonComponent changes.
- I combined the select and deselect functions in the Aggregation Subcomponents.
- I replaced the need to override handleClickEvent and handleHoverEvent with a selectMode property in the Aggregation Subcomponents.
- I moved a lot of the complex select/deselect behavior to other tickets and left TODOs with the ticket names.


src/app/components/aggregation/aggregation.component.html
```diff
@@ -20,38 +20,16 @@
         <mat-spinner></mat-spinner>
     </div>
 
-    <div class="filter-container" *ngIf="showHeaderContainer()">
-        <app-legend *ngIf="showLegend()" [activeList]="legendActiveGroups" [colorKeys]="colorKeys"
+    <div class="filter-container" *ngIf="showLegend()">
```
Konnoke, line 23: I agree with moving the ngif from the app-legend to the div element because it would persist the legend component instead of creating and destroying the legend component each time but are there other benefits?

Konnoke (approved):

src/app/components/aggregation/aggregation.component.html
```diff
@@ -71,12 +49,4 @@
     </div>
 
     <canvas #hiddenCanvas class="hidden-canvas"></canvas>
-
```
rartiss55: It looks like space for the contribution footer (THOR-305) is still being made for the aggregation component, but that the display logic here was removed. I also noticed the space/contribution text disappeared completely from the data table. Please put this logic back in for components that it was removed from (should appear if needed for all components). Various LORELEI configs you can use to test include:
- TransitionPartner
- APAN
- HumanityRoads

rartiss55 (changes requested):
See earlier comment.

src/app/components/aggregation/aggregation.component.html
ThomasSchellenbergNextCentury: This change wasn't as witty as you might think.  Previously, the "filter-container" contained both the legend and the gray filter boxes.  The showHeaderContainer function returned whether either of those needed to be visible.  By removing the gray filter boxes, the only thing that needs to be shown is the legend, so I replaced showHeaderContainer with showLegend.  Does that answer your question?

src/app/components/aggregation/aggregation.component.html
ThomasSchellenbergNextCentury: Ugh, OK, I thought I had merged it in correctly, but obviously I messed something up.  I'll fix the aggregation component and data table, and I'll also check the other components to make sure they have the contribution footers too.

rartiss55 (approved):

PR #187, opened 2019-04-19
THOR-942 Filter Refactor (Filter Builder)

ThomasSchellenbergNextCentury:

HOW DOES THOR-942 FIX IT

WHAT IS IN THIS PR
- Updates to the Aggregation Component to make it work with the new FilterService and BaseNeonComponent changes.
- I had to change the core behavior of the Filter Builder to make it work in the refactor.  Since individual components no longer have ownership of their own filters, it doesn't make sense for the Filter Builder to hold onto or show only the filters it created.  Therefore the workflow for the Filter Builder (for now) is as follows:  the user designs a filter using the dropdowns and input fields as normal; adding clauses to the filter builder transforms the design into a compound AND/OR filter; clicking "Save Filter" adds a single filter, created using all valid clauses, to the dashboard, and resets the Filter Builder; existing filters can be deleted using the Active Filters tab, like filters from visualizations.  I'm going to review this new workflow with Sophie next week so we can plan for our next steps.

### DIFF
--- a/src/app/components/filter-builder/filter-builder.component.ts
+++ b/src/app/components/filter-builder/filter-builder.component.ts
@@ -62,7 +62,9 @@ export class FilterBuilderComponent extends BaseNeonComponent implements OnInit,
         { value: '>=', prettyName: '>=' },
         { value: '<=', prettyName: '<=' }
     ];
-    public requireAll: boolean = false;
+
+    public compoundTypeIsOr: boolean = false;
+    public filterIsOptional: boolean = false;
 
     constructor(
         datasetService: DatasetService,
@@ -262,12 +264,12 @@ export class FilterBuilderComponent extends BaseNeonComponent implements OnInit,
      * Saves a new custom filter using every filter clause in the global list.
      */
     public saveFilter(): void {
-        if (!this.validateFilters(this.filterClauses)) {
+        if (!this.filterClauses.length || !this.validateFilters(this.filterClauses)) {
             return;
         }
 
         // Turn the filter clauses into filter designs.
-        let filterDesigns: FilterDesign[] = this.filterClauses.map((filterClause) => {
+        let filterDesigns: SimpleFilterDesign[] = this.filterClauses.map((filterClause) => {
             let operator: string = filterClause.operator.value;
             let value: any = filterClause.value;
             if (filterClause.operator.value !== 'contains' && filterClause.operator.value !== 'not contains') {
@@ -278,6 +280,7 @@ export class FilterBuilderComponent extends BaseNeonComponent implements OnInit,
                 }
             }
             return {
+                optional: this.filterIsOptional,
                 datastore: '',
                 database: filterClause.database,
                 table: filterClause.table,
@@ -289,7 +292,8 @@ export class FilterBuilderComponent extends BaseNeonComponent implements OnInit,
 
         // Create a compound filter from multiple filters if needed.
         let filterDesign: FilterDesign = !filterDesigns.length ? null : (filterDesigns.length === 1 ? filterDesigns[0] : {
-            type: this.requireAll ? CompoundFilterType.AND : CompoundFilterType.OR,
+            type: this.compoundTypeIsOr ? CompoundFilterType.OR : CompoundFilterType.AND,
+            optional: this.filterIsOptional,
             inflexible: true,
             filters: filterDesigns
         } as CompoundFilterDesign);
--- a/src/app/components/filter-builder/filter-builder.component.ts
+++ b/src/app/components/filter-builder/filter-builder.component.ts
@@ -23,9 +23,9 @@ import {
     ViewEncapsulation
 } from '@angular/core';
 
-import { AbstractSearchService, FilterClause, QueryPayload } from '../../services/abstract.search.service';
+import { AbstractSearchService, CompoundFilterType, FilterClause, QueryPayload } from '../../services/abstract.search.service';
 import { DatasetService } from '../../services/dataset.service';
-import { FilterService } from '../../services/filter.service';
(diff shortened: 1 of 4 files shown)


Konnoke (approved):

src/app/components/filter-builder/filter-builder.component.ts
```diff
@@ -81,91 +84,47 @@ export class FilterBuilderComponent extends BaseNeonComponent implements OnInit,
...
+        let filterClause: FilterClauseMetaData = new FilterClauseMetaData(() => []);
+        filterClause.updateDatabases(this.datasetService);
+        filterClause.field = this.createEmptyField();
+        filterClause.operator = this.operators[0];
+        filterClause.value = '';
+        filterClause.changeDatabase = filterClause.database;
+        filterClause.changeTable = filterClause.table;
+        filterClause.changeField = filterClause.field;
+        if (filterClause.database && filterClause.table) {
+            this.filterClauses.push(filterClause);
```
gitnordt, line 105: When adding multiple OR filters, I found myself wishing that I did not have to keep setting the database and field. Would it be difficult to capture the previous filter's values and add them to the new filter for convenience? It would be less work on the user's part.

src/app/components/filter-builder/filter-builder.component.ts
```diff
@@ -461,96 +241,69 @@ export class FilterBuilderComponent extends BaseNeonComponent implements OnInit,
...
-    setupFilters() {
-        // Do nothing.
+        // Create a compound filter from multiple filters if needed.
+        let filterDesign: FilterDesign = !filterDesigns.length ? null : (filterDesigns.length === 1 ? filterDesigns[0] : {
+            type: this.compoundTypeIsOr ? CompoundFilterType.OR : CompoundFilterType.AND,
+            optional: this.filterIsOptional,
+            inflexible: true,
+            filters: filterDesigns
+        } as CompoundFilterDesign);
+
```
gitnordt, line 300: What is the expected behavior for adding optional filters that contain an AND compound filter?

Query 1: 
optional (Network / Seedling / categories = ENTITY) or (Network / Seedling / categories = RELATION)
optional Network / Seedling / categories = EVENT

Query 1 Results: Shows all 3 categories(ENTITY, RELATION, and EVENT) as expected.

Query 2:
optional (Network / Seedling / categories = ENTITY) or (Network / Seedling / categories = RELATION)
optional (Network / Seedling / categories = EVENT) and (Network / Seedling / name contains soldier)

Query 2 Results: No results returned. My expectation was to see all items in the ENTITY and RELATION categories and only items in the EVENT categories that contain 'soldier' in the name field. 

Is my expectation incorrect?

src/app/components/filter-builder/filter-builder.component.ts
ThomasSchellenbergNextCentury: Good idea!  I'll investigate how easy this is and either do it or make a ticket for it.

src/app/components/filter-builder/filter-builder.component.ts
ThomasSchellenbergNextCentury: I can duplicate the issue.

Your expectation doesn't match the current intended behavior, but maybe the behavior is wrong and your expectation is correct...

Let me think about this and get back to you tomorrow.

ThomasSchellenbergNextCentury: @snorpet Oh, OK, I remember why I wanted it to behave this way.

So, here's what's happening:

If you have a single "optional" filter on a field/set-of-fields ("FilterDataSource"), then all the query results must match that single filter.  If you have multiple "optional" filters, then the query results must match at least one of them.  This you know.

So if you have `optionalFilter1` and `optionalFilter2` then they're combined into `OR( [ optionalFilter1, optionalFilter2 ] )`.

The confusion comes from the fact that these two filters you've created are on two DIFFERENT FilterDataSources:  the first is on `[categories=, categories=]` and the second is on `[categories=, name contains]`.  This means that the filters, while both "optional", are kept separate and added to the query like this:

```
AND(
  [
    OR(
      [
        OR(
          [
            categories=ENTITY,
            categories=RELATION
          ]
        )
      ]
    ),
    OR(
      [
        OR (
          [
            categories=EVENT,
            name contains soldier
          ]
        )
      ]
    )
  ]
)
```

What you want to do is use the Filter Builder to create a SINGLE filter (not two filters) with multiple levels of nesting:

```
OR(
  [
    OR(
      [
        categories=ENTITY,
        categories=RELATION
      ]
    ),
    OR (
      [
        categories=EVENT,
        name contains soldier
      ]
    )
  ]
)
```

Unfortunately, you can't make filters with multiple levels of nesting right now in the Filter Builder, but there's a ticket to implement it (https://nextcentury.atlassian.net/browse/THOR-855).

My takeaways from this:
- I think the current behavior is correct, considering how the filters from the filter builder interact with the filters from the visualizations, but we can go over it in person tomorrow.
- Assuming this is true, we need a better way of communicating how it works to the user.  I can definitely understand how you arrived at your expectation given the current workflow (reviewing it confused me too!).  I think that would involve changing some of the terminology (like "required" and "optional"), but I'll need to brainstorm a bit more on it tonight.

gitnordt: Great explanation, thank you for clarifying. This also ties into our conversation a couple of days ago about the terminology. No rush on brainstorming, keep me posted if you come up with anything and I will do the same.

gitnordt (approved):

src/app/components/filter-builder/filter-builder.component.ts
ThomasSchellenbergNextCentury: Done!  Check it out on the THOR-942-services branch:  https://github.com/NextCenturyCorporation/neon-dash-internal/commit/502cc5d87b97367c8c6ca03964b06a3ccb74f2e6

PR #188, opened 2019-04-19
THOR-942 Filter Refactor (Query Bar and Taxonomy)

ThomasSchellenbergNextCentury:

HOW DOES THOR-942 FIX IT

WHAT IS IN THIS PR
- Updates to the Query Bar and Taxonomy Viewer Components to make them work with the new FilterService and BaseNeonComponent changes.
- I made a big change in how the Taxonomy specifically creates filters:  rather than adding or removing filters on individual unchecked/checked nodes and their children/parents, it now completely replaces its previous filters with new filters representing ALL unchecked nodes (and their children/parents) whenever a node is unchecked/checked.  I think I tested this new behavior pretty well, but, Stacey, I want to get your feedback to make sure I haven't made any mistakes!

### DIFF
--- a/src/app/components/query-bar/query-bar.component.ts
+++ b/src/app/components/query-bar/query-bar.component.ts
@@ -13,18 +13,18 @@
  * limitations under the License.
  *
  */
-import { BehaviorSubject, Observable } from 'rxjs';
+import { Observable } from 'rxjs';
 import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ElementRef, Injector, ViewChild } from '@angular/core';
 import { FormControl } from '@angular/forms';
 import { map, startWith } from 'rxjs/operators';
 
-import { AbstractSearchService, FilterClause, QueryPayload } from '../../services/abstract.search.service';
+import { AbstractSearchService, CompoundFilterType, FilterClause, QueryPayload } from '../../services/abstract.search.service';
 import { AbstractWidgetService } from '../../services/abstract.widget.service';
 import { DatasetService } from '../../services/dataset.service';
-import { FilterService } from '../../services/filter.service';
+import { CompoundFilterDesign, FilterBehavior, FilterDesign, FilterService, SimpleFilterDesign } from '../../services/filter.service';
 
 import { BaseNeonComponent, TransformedVisualizationData } from '../base-neon-component/base-neon.component';
-import { FieldMetaData, SimpleFilter } from '../../dataset';
+import { DatabaseMetaData, FieldMetaData, TableMetaData } from '../../dataset';
 import { neonUtilities } from '../../neon-namespaces';
 import {
     OptionChoices,
@@ -54,11 +54,7 @@ export class QueryBarComponent  extends BaseNeonComponent {
     autoComplete: boolean = true;
     queryValues: string[] = [];
     queryArray: any[];
-    filterIds: string[] = [];
-    currentFilter: string = '';
 
-    public simpleFilter = new BehaviorSubject<SimpleFilter>(undefined);
-    public filterId = new BehaviorSubject<string>(undefined);
     public queryOptions: Observable<void | string[]>;
 
     private filterFormControl: FormControl;
@@ -98,6 +94,43 @@ export class QueryBarComponent  extends BaseNeonComponent {
         ];
     }
 
+    private createFilterDesignOnExtensionField(
+        databaseName: string,
+        tableName: string,
+        fieldName: string,
+        value?: any
+    ): FilterDesign {
+        let database: DatabaseMetaData = this.datasetService.getDatabaseWithName(databaseName);
+        let table: TableMetaData = this.datasetService.getTableWithName(databaseName, tableName);
+        let field: FieldMetaData = this.datasetService.getFieldWithName(databaseName, tableName, fieldName);
+        return (database && database.name && table && table.name && field && field.columnName) ? {
+            datastore: '',
+            database: database,
+            table: table,
+            field: field,
+            operator: '=',
+            value: value
+        } as SimpleFilterDesign : null;
+    }
+
+    private createFilterDesignOnList(filters: FilterDesign[]): FilterDesign {
+        return {
+            type: CompoundFilterType.OR,
+            filters: filters
+        } as CompoundFilterDesign;
+    }
+
+    private createFilterDesignOnText(value?: any): FilterDesign {
+        return {
+            datastore: '',
+            database: this.options.database,
+            table: this.options.table,
+            field: this.options.filterField,
+            operator: '=',
+            value: value
+        } as SimpleFilterDesign;
+    }
+
     /**
      * Creates and returns an array of non-field options for the visualization.
      *
@@ -112,6 +145,45 @@ export class QueryBarComponent  extends BaseNeonComponent {
             new WidgetFreeTextOption('id', 'ID', ''),
             new WidgetFreeTextOption('placeHolder', 'Place Holder', 'Query')
         ];
+    }
+
+    /**
+     * Returns each type of filter made by this visualization as an object containing 1) a filter design with undefined values and 2) a
+     * callback to redraw the filter.  This visualization will automatically update with compatible filters that were set externally.
+     *
+     * @return {FilterBehavior[]}
+     * @override
+     */
+    protected designEachFilterWithNoValues(): FilterBehavior[] {
+        let behaviors: FilterBehavior[] = [];
+
+        if (this.options.filterField.columnName) {
+            behaviors.push({
+                // Match a single EQUALS filter on the filter field.
+                filterDesign: this.createFilterDesignOnText(),
+                redrawCallback: this.updateQueryBarText.bind(this)
+            });
+        }
+
+        if (this.options.extendedFilter) {
+            this.options.extensionFields.forEach((extensionField) => {
+                behaviors.push({
+                    // Match a single EQUALS filter on the extension database/table/field.
+                    filterDesign: this.createFilterDesignOnExtensionField(extensionField.database, extensionField.table,
+                        extensionField.idField),
+                    redrawCallback: () => { /* Do nothing */ }
+                });
+
+                behaviors.push({
+                    // Match a compound OR filter with one or more EQUALS filters on the extension database/table/field.
+                    filterDesign: this.createFilterDesignOnList([this.createFilterDesignOnExtensionField(extensionField.database,
+                        extensionField.table, extensionField.idField)]),
+                    redrawCallback: () => { /* Do nothing */ }
+                });
+            });
+        }
+
+        return behaviors;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -528,42 +472,77 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
+        //     this.createFilterDesignOnList(group.sourceIds.map((sourceId) =>
+        //         this.createFilterDesign(this.options.sourceIdField, sourceId))));
+        // let filterOnSource = !!(sourceFilters.length);
+        // let sourceFilter: FilterDesign = filterOnSource ? (sourceFilters.length === 1 ? sourceFilters[0] :
+        //     this.createFilterDesignOnList(sourceFilters)) : null;
+        // if (!sourceFilter && this.options.sourceIdField.columnName) {
+        //   filterDesignListToDelete.push(this.createFilterDesign(this.options.sourceIdField));
+        // }
+
+        this.exchangeFilters([categoryFilter, typeFilter, subTypeFilter].filter((filter) => !!filter), filterDesignListToDelete);
```
gitnordt, line 545: The taxonomy checkbox behavior: 

- When all child nodes are checked, the parent should also be checked. 

- When a parent node has at least one child that is indeterminate, the parent should also be indeterminate

- When a child node is unchecked in the taxonomy and it is the only child or all of its siblings are unchecked, it is expected that the parent node is unchecked. 

- When a child node is unchecked and it has siblings that are checked, it is expected that the parent node is indeterminate (-).  

updateChildNodesCheckbox and updateParentNodesCheckBox are the methods responsible for this functionality. However, it looks like they may be getting overriden by a call to refreshVisualization in base-neon.component.  I am seeing the parent node change to indeterminate and then, shortly after, change to checked. Putting the update checkbox methods after the exchangeFilters call may fix the issue, but I'm not certain.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -528,42 +472,77 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
-
-        //Gather the top level nodes in the taxonomy that are unchecked and add a != filter
-        if (node.parent.level === 0 && $event.target.checked === false) {
-            this.createNodeFilter(node.data, []);
-        } else if (node.parent.level > 0) {
-            //Add parents' siblings if they exist
-            if (node.parent.parent && !node.parent.parent.data.virtual) {
-                relatives = this.retrieveUnselectedNodes(node.parent.parent.data.children);
-            }
+        // Find all the unselected groups in the taxonomy (parents and children).
```
gitnordt, line 487: I am seeing a strange one-off bug. When using AIDA data for the taxonomy, if you open the first category which is ENTITY and uncheck the GeneralAffiliation child node, you can never check it again. None of the other categories have this problem from what I can see. I think it is safe to say it is specific to the first subType object created.

gitnordt (changes requested):

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -134,11 +159,51 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
+            //     filterDesign: this.createFilterDesignOnList([this.createFilterDesign(this.options.sourceIdField)]),
+            //     redrawCallback: () => {}
+            // });
+            // behaviors.push({
+            //     // Match a compound AND filter with one or more compound AND filters with one or more NOT EQUALS filters.
+            //     filterDesign: this.createFilterDesignOnList([this.createFilterDesignOnList(
+            //         [this.createFilterDesign(this.options.sourceIdField)])]),
+            //     redrawCallback: () => {}
+            // });
+        }
```
gitnordt, line 202: Do you have an ETA on when we can expect filtering for the sourceIdField to be implemented? I ask because we did have a user point out that filtering was not working for documents in the taxonomy view.

src/app/components/query-bar/query-bar.component.ts
```diff
@@ -239,82 +311,48 @@ export class QueryBarComponent  extends BaseNeonComponent {
...
+        let removeTextFilter = true;
+        let removeFilterList: FilterDesign[] = [];
 
-        //filters query text
-        if (text && text !== this.currentFilter) {
-            let values = this.queryArray.filter((value) =>
-                value[this.options.filterField.columnName].toLowerCase() === text.toLowerCase()),
-                clause: WherePredicate;
+        if (text) {
+            let values: any[] = this.queryArray.filter((value) =>
```
gitnordt, line 324: I am not getting any values for this.queryArray. I am seeing the following error in the console:

QueryBarComponent.html:6 ERROR TypeError: Cannot read property 'filter' of undefined
    at QueryBarComponent.push../src/app/components/query-bar/query-bar.component.ts.QueryBarComponent.createFilter (query-bar.component.ts:327)
    at Object.eval [as handleEvent] (QueryBarComponent.html:7)
    at handleEvent (core.js:21652)
    at callWithDebugContext (core.js:22746)
    at Object.debugHandleEvent [as handleEvent] (core.js:22449)
    at dispatchEvent (core.js:19101)
    at core.js:19548
    at HTMLInputElement.<anonymous> (platform-browser.js:993)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invokeTask (zone.js:423)
    at Object.onInvokeTask (core.js:16126)

gitnordt (changes requested):

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
ThomasSchellenbergNextCentury: This is what Scott's working on right now (https://nextcentury.atlassian.net/browse/NEON-22).  It really depends on how quickly he can get the first part done (the server-specific part), and then how quickly we can integrate it with the rest of the system (at least that can be divided among multiple people).  I'd say 2-4 weeks, but that's a rough guess.  Does that help, at least?

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
ThomasSchellenbergNextCentury: Great catch!  I was able to reproduce your issue.

I think this was not happening previously because the taxonomyViewer usually sets ignoreSelf to true.  However, there was a bug in the BaseNeonComponent causing the taxonomyViewer to requery itself (and thus call refreshVisualization) even when ignoreSelf equals true.  I've fixed the bug and can confirm that the behaviors you want actually happen now.

Side note:  Is there ever a situation in which we'd want ignoreSelf to be false for the taxonomyViewer?  If not, the taxonomyViewer can override the shouldFilterSelf function to always return false.  That might be useful so that this same problem doesn't occur if we forget to set ignoreSelf to true on the taxonomyViewer.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
ThomasSchellenbergNextCentury: I was able to reproduce this issue.  However, it must have also been caused by the calls to refreshVisualization, because my solution to your previous comment also solved this issue too.  Yay!

src/app/components/query-bar/query-bar.component.ts
ThomasSchellenbergNextCentury: Wow, how did I let THAT slip past me?!  My bad.

Looks like initializing the queryArray fixes the issue.

Side note:  I found some funny behavior with the extensions filters that I think (maybe) is a config issue.  I'll stop by tomorrow to talk to you about it.

ThomasSchellenbergNextCentury:
@snorpet The bug fixes to your comments above are on the THOR-942-services branch here:  https://github.com/NextCenturyCorporation/neon-dash-internal/commit/05a5d213983102be15d2447eab1acfce921176bf

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
gitnordt: I don't believe there is a circumstance where we would want ignoreSelf to be false. That would mean that as you check boxes, values in the taxonomy would be filtered out. I think that would be very confusing for the user.

src/app/components/query-bar/query-bar.component.ts
gitnordt: Just wondering are the pre-populated query values coming up for you as you start to type a question in the query bar? 

I am getting a different error in the console now:

ERROR TypeError: Cannot read property 'datastore' of null
    at Object.isSimpleFilterDesign (filter.service.ts:212)
    at Object.createFilterDataSourceListFromDesign (filter.service.ts:114)
    at SingleListFilterCollection.push../src/app/services/filter.service.ts.AbstractFilterCollection.findFilterDataSources (filter.service.ts:237)
    at filter.service.ts:918
    at Array.forEach (<anonymous>)
    at FilterService.push../src/app/services/filter.service.ts.FilterService.updateCollectionWithGlobalCompatibleFilters (filter.service.ts:916)
    at QueryBarComponent.push../src/app/components/base-neon-component/base-neon.component.ts.BaseNeonComponent.updateCollectionWithGlobalCompatibleFilters (base-neon.component.ts:790)
    at QueryBarComponent.push../src/app/components/base-neon-component/base-neon.component.ts.BaseNeonComponent.ngOnInit (base-neon.component.ts:174)
    at checkAndUpdateDirectiveInline (core.js:20644)
    at checkAndUpdateNodeInline (core.js:21908)

gitnordt (changes requested):
I'm still having issues with the query bar.

src/app/components/query-bar/query-bar.component.ts
ThomasSchellenbergNextCentury: @snorpet The extensionFields bindings for the queryBar in your configFile are incorrect, assuming that this is your config file:  https://github.com/NextCenturyCorporation/neon-dash-internal/blob/verdi-master/src/app/config/config.verdi-ta2-ta3.yaml#L647

The extensionFields format is the same as before; it doesn't use table keys or field keys.  Since the extensionFields are a custom binding defined by the Query Bar, the changes that affected the other fields don't affect it, so someone would need to change the custom code in the Query Bar to update it to use table keys and field keys.

Keep it like this:
```
extensionFields:
  - database: verdi-support
    table: hypothesis_info
    idField: hypothesis_id
    filterField: query
  - database: kb-simple
    table: seedling
    idField: docIds
    filterField: hypotheses
```

ThomasSchellenbergNextCentury: Also:  I think I fixed my issue from before.  Please pull the latest code from THOR-942-services.

https://github.com/NextCenturyCorporation/neon-dash-internal/commit/f9f5bfa04e3e1ba7815f0f2483dda4b68b5ba188

gitnordt: Thank you, I will try testing with these extensionField values.

gitnordt (approved):

PR #189, opened 2019-04-19
THOR-942 Filter Refactor (Supplemental Files)

ThomasSchellenbergNextCentury:

HOW DOES THOR-942 FIX IT

WHAT IS IN THIS PR
- Updates to the non-filtering subclasses of BaseNeonComponent and other affected non-visualization components to make them work with the new FilterService and BaseNeonComponent changes.


src/app/app.component.ts
```diff
@@ -261,12 +261,8 @@ export class AppComponent implements AfterViewInit, OnInit, OnDestroy {
...
 
     changeFiltersComponentIcon() {
-        let filters = this.filterService.getFilters();
-        if (filters.length > 0) {
-            this.filtersIcon = 'filters_active';
-        } else {
-            this.filtersIcon = 'filters';
-        }
+        this.filtersIcon = this.isFiltered() ? 'filters_active' : 'filters';
+        // TODO Does this function really have to return a boolean value?
```
Konnoke, line 265: It's better to return a boolean value for this function because it's being used for an ngIf

Konnoke (approved):

PR #190, opened 2019-04-19
THOR-942 Filter Refactor (BaseNeonComponent and Subclasses)

ThomasSchellenbergNextCentury:

HOW DOES THOR-942 FIX IT

WHAT IS IN THIS PR
- Changes to the BaseNeonComponent itself, as detailed above.
- Updates to the filtering subclasses of BaseNeonComponent (except the Query Bar and Taxonomy Viewer, because they have their own PR) to make them work with the new FilterService and BaseNeonComponent changes.


src/app/components/annotation-viewer/annotation-viewer.component.ts
```diff
@@ -200,32 +229,16 @@ export class AnnotationViewerComponent extends BaseNeonComponent implements OnIn
...
-            prettyField: part.type
-        };
-        if (this.filterIsUnique(filter) && part.annotation) {
-            this.addLocalFilter(filter);
-            let whereClause = neon.query.where(filter.field, '=', filter.value);
-            this.addNeonFilter(this.options, true, filter, whereClause);
-        }
-        if (!part.annotation) {
+        if (part.annotation) {
+            // TODO THOR-1098
```
rartiss55, line 239: Re: THOR-1098, what did this do before/what needs to be fixed?

src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -816,30 +766,47 @@ export abstract class BaseNeonComponent implements AfterViewInit, OnInit, OnDest
...
+            // TODO THOR-1108 Ignore filters on non-matching datastores/databases/tables.
+            this.executeAllQueryChain();
+        }
     }
 
     /**
-     * Get and configure filters from the filter service.
-     * DO NOT EXECUTE QUERIES IN THIS METHOD.
-     * This method will be called before the visualization has finished initialization!
+     * Replaces the existing global widget filters with all the compatible (matching) dash filters.
```
rartiss55, line 786: Do we mean replacing global widget filters with what filters apply specifically for this visualization? Or am I reading this backwards?

src/app/components/annotation-viewer/annotation-viewer.component.ts
ThomasSchellenbergNextCentury: I'm not really sure what this did before.  It let you set filters on annotation "parts", whatever that means.  We're not using it now and I have no way of testing it (even if I understood what it did).  Figuring all that out seemed outside the scope of the ticket which is why I left the TODO.

src/app/components/base-neon-component/base-neon.component.ts
ThomasSchellenbergNextCentury: Yes:  Anything that's set in this widget's collection of filters will be replaced by the filters that apply specifically to this widget/visualization (a.k.a. the "compatible" filters).  Does the comment need clarification?

rartiss55: Yeah I think some clarification here would help.

rartiss55 (changes requested):
Left an earlier note to clarify something in the base-neon.component, but otherwise I think everything looks good and I tested with a couple different LORELEI datasets. The TODOs listed don't seem like high priority at the moment, and they're pretty easy to find under the 'Filter Changes' epic.

@bdpham I only checked the widgets we use for LORELEI, so wanted to leave a note for you to be sure to test out the changes made in the network graph, thumbnail grid, newsfeed and timeline.

PR #191, opened 2019-04-19
THOR-942 Filter Refactor (FilterService and DatasetService)

ThomasSchellenbergNextCentury:

HOW DOES THOR-942 FIX IT

WHAT IS IN THIS PR
- Changes to the FilterService, as detailed above.
- Updates to how relations are handled in the DatasetService.


src/app/services/dataset.service.spec.ts
```diff
@@ -87,29 +87,206 @@ describe('Service: mock DatasetService with mock data', () => {
...
+                [
+                    ['datastore1.testDatabase1.fakeTable1.fakeRelationFieldA', 'datastore1.testDatabase1.fakeTable1.fakeRelationFieldA'],
+                    ['datastore1.testDatabase2.testTable2.testRelationFieldA', 'datastore1.testDatabase2.testTable2.testRelationFieldB']
+                ],
+                [
+                    ['datastore1.testDatabase1.testTable1.fakeRelationFieldA', 'datastore1.testDatabase1.testTable1.fakeRelationFieldA'],
+                    ['datastore1.testDatabase2.testTable2.testRelationFieldA', 'datastore1.testDatabase2.testTable2.testRelationFieldB']
+                ]
+            ]
+        });
```
gitnordt, line 269: Can you elaborate on why this is considered "bad databases/tables/fields" ? Is it because you are passing the fake databases/tables/fields?

src/app/services/dataset.service.ts
```diff
@@ -1096,6 +976,47 @@ export class DatasetService {
...
+     * the inner array are specific fields within the relations.
+     *
+     * @return {SingleField[][][]}
+     */
+    public findRelationDataList(): SingleField[][][] {
+        let configRelationDataList: (string | string[])[][] = this.getCurrentDashboard().relations || [];
+
+        // Either expect string list structure:  [[a1, a2, a3], [b1, b2]]
+        // ....Or expect nested list structure:  [[[x1, y1], [x2, y2], [x3, y3]], [[z1], [z2]]]
+        return configRelationDataList.map((configRelationData) => {
```
gitnordt, line 1010: I don't fully understand the need for a triple nested list structure. Maybe this can be better explained in-person.

src/testUtils/MockServices/FilterServiceMock.ts
```diff
@@ -13,66 +13,8 @@
...
-        id: string,
-        onSuccess?: (resp: any) => any,
-        onError?: (resp: any) => any) {
-
-        let index = _.findIndex(this.getFilters(), {id: id});
-        this.getFilters().splice(index, 1);
-
-        // don't do success call to avoid calling query chain
-    }
+    // TODO
```
gitnordt: Can you refresh my memory of why we are keeping this file around?

src/app/services/dataset.service.ts
```diff
@@ -1096,6 +976,47 @@ export class DatasetService {
...
+            return configRelationData.map((configRelationFilterFields) => {
+                // A relation is an array of arrays.  The elements in the outer array are the fields-to-substitute and the elements in the
+                // inner arrays are the filtered fields.  The inner arrays must be the same length (the same number of filtered fields).
+                let relationFilterFields: string[] = Array.isArray(configRelationFilterFields) ? configRelationFilterFields :
+                    [configRelationFilterFields];
```
gitnordt, line 1015: Something to think about: Is there an advantage to using Array.isArray vs. instanceof Array? If so, we may want to make it consistent throughout our code.

src/app/services/dataset.service.spec.ts
rartiss55: To add to this, @ThomasSchellenbergNextCentury could you update the relations part for the darpa-July2018/darpa-July2018-es6 config files? It might help to have a working example in an existing config we can reference.

src/app/services/dataset.service.ts
```diff
@@ -751,173 +798,6 @@ export class DatasetService {
...
-
-    public getEquivalentFields(datastore: string,
-        database: string,
-        table: string,
-        field: string,
-        mapping: Map<string, Map<string, { datastore: string, database: string, table: string, field: string }[]>>):
-        Map<string, Map<string, { datastore: string, database: string, table: string, field: string }[]>> {
-        let relatedFields: any = mapping;
-
-        // TODO: THOR-1063: filters will need datastore info eventually - more of a backend task,
```
rartiss55, line 795: Is there a new place we can leave this TODO so we don't forget about this issue? Or does this not matter anymore and we should delete the ticket?

src/app/services/filter.service.ts
```diff
@@ -14,417 +14,1259 @@
...
+    table: TableMetaData;
+    field: FieldMetaData;
+    operator: string;
+    value?: any;
+}
 
-    constructor(protected datasetService: DatasetService) {}
+export interface CompoundFilterDesign extends FilterDesign {
+    type: CompoundFilterType;
+    inflexible?: boolean;
```
rartiss55: I read what's in isCompatibleWithDesign() but I'm still having a bit of trouble with what inflexible means here exactly? Is it enforcing a specific order on nested filters?

src/app/services/filter.service.ts
```diff
@@ -14,417 +14,1259 @@
...
+            console.error('DualListFilterCollection has bad data!', this.data);
+            return [];
         }
+        return filters[0] ? (filters[0] as CompoundFilter).filters : [];
+    }
 
-        this.messenger.publish(neonEvents.FILTERS_CHANGED, {});
+    /**
+     * Returns the nested required and optional filters for the given data source (or an existing matching data source within this
+     * collection) as a single filter list.
```
gitnordt: Is required and optional filters pertaining to required and optional fields or does this have to do with "and/or" filters?

rartiss55:
I think it'd be good to document the filter.service – its probably the one thing that is kind of a black box to people who haven’t worked on it and not as straightforward as the other services because of its complexity. Here are some starting points:
-	Decision making for how its structured (you addressed some of this in your talk/on this PR, but it’d be helpful to have that documented for the future)
-	The specific structure of things (for instance, we were a bit confused why the FilterDataSource interface is needed since it is just a list of FilterClauseDataSources, but I’m sure there’s a good reason :))
-	Terminology (i.e. meaning of inflexible, required vs optional filtering for clarification)
-	In general summaries of what happens on filter changes/what events get called and why

src/app/services/dataset.service.spec.ts
ThomasSchellenbergNextCentury: @snorpet Ah, sorry for the confusion.  Yes, by "bad" I meant fake or non-existent.  I'll change "bad" to be clearer.

@rartiss55 I'll do that too!

src/app/services/dataset.service.ts
ThomasSchellenbergNextCentury: @snorpet We can definitely talk about it in-person, but here's the summary:

- Each element in the first (outermost) list is a separate relation.
- Each element in the second list is a related field (or set of related fields).
- Each element in the third (innermost) list is one of the fields that's expected in the filter.  A filter must have exactly all of the fields within this list in order for this relation to be applied.

Example:
```javascript
[ // relation list
  [ // single relation
    [ // filter fields
      A,
      B
    ],
    [ // filter fields
      X,
      Y
    ]
  ]
]
```

In this example, whenever a filter contains exactly A and B, they are replaced by X and Y (A => X and B => Y), and visa-versa.  If a filter contains just A, or A and B and C, then the relation would not be applied.

Does that make sense?

src/testUtils/MockServices/FilterServiceMock.ts
ThomasSchellenbergNextCentury: We actually probably don't need it any more, so I should delete it.  (It was originally there to override calls to the server during unit tests, but the FilterService doesn't talk to the server any more.)

src/app/services/dataset.service.ts
ThomasSchellenbergNextCentury: So, I hadn't thought about it before, but I did a bit of research, and it looks like, these days, Array.isArray is preferred because it works through iframes.  Here's a recent article:  https://codewithhugo.com/detecting-object-vs-array-in-javascript-by-example/

While I doubt the iframe issue will be relevant, it still seems that this reason gives Array.isArray an edge over instanceof.  You're also right that it would be nice to make it consistent throughout the code; I'll add a note in my "new lint rules to consider" list.

src/app/services/dataset.service.ts
ThomasSchellenbergNextCentury: The THOR-1063 ticket has been converted to NEON-24 and is one of the things Scott is working on.  I think that ticket is more of a backend task and the corresponding frontend ticket is THOR-1078 which is littered in comments throughout the code.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: There's some more information in the handouts I gave you this morning, but to summarize:

- Flexible is for compound filters that can have one or more nested filters.  For example, by clicking bars on a bar chart, you set one or more "equals" filters on the bar field.
- Inflexible is for compound filters that have a specific nested filter configuration.  For example, the bounding box filter in the map must have exactly four filters:  >= lat field, <= lat field, >= lon field, <= long field.

I'll add some additional comments in the code.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: Ah, yeah, this was some new behavior that I forgot to review with you all -- silly me.

Currently, visualizations like the bar chart, data table, and text cloud have a filter operator config property that determines the compound filter type (AND/OR) when that visualization sets multiple filters (like multiple bars in the bar chart or multiple items in the text cloud).

Separating filters that should be ANDed together from filters that should be ORed together created complications so I added the "optional" property to the filters.  By default, all filters are required, but visualizations should make filters optional when they want the filters to be ORed together.

When you have one or more "optional" filters on a field, the search must satisfy at least one of those filters.  And "optional" filters on different fields (with different FilterDataSources) work independently, creating separate OR filters on each individual field.

The required/optional terminology is what we used on Memex, and I think it makes sense (once you know the details), but I've also asked Sophie to review the terminology, in case there's a better alternative.

Sorry for the confusion.  Hopefully this helps clear things up.  I'll add some additional comments in here somewhere too.

ThomasSchellenbergNextCentury:
@rartiss55 Concerning the FilterDataSource vs. FilterClauseDataSources:  I think you're right that we can probably just change FilterDataSource into FilterClauseDataSource[].

For a bit of (somewhat excessive) background...

I think the more-complicated object structure came from the fact that I was trying to make non-primitive keys in a Map work, and I thought that maybe an Object would work better than an Array (this was before I made the FilterCollection object).  Unfortunately, both Objects and Arrays ended up having the same problem:  if you use one as a key in a Map, you need to use that exact variable to access the value from the Map, and not simply an equivalent Object/Array with exactly the same properties.  This was why I had to create the FilterCollection object so it could manage equivalent FilterDataSource keys.

I don't think the extra complexity in the FilterDataSource is needed any more, and making it simpler seems like a good idea, so I'll do that.

ThomasSchellenbergNextCentury:
Finished with all of the CR changes requested so far.

src/app/services/dataset.service.ts
gitnordt: Good idea and very informative article. Array.isArray definitely wins as far as reliability.

gitnordt (dismissed):

rartiss55 (approved):

gitnordt (approved):

PR #194, opened 2019-04-29
THOR-877 Moved Behavior to Load Dashboard State

ThomasSchellenbergNextCentury:
Moved behavior to load dashboard state from the dashboard selector component into the app component and dataset service.

### DIFF
--- a/src/app/dataset.ts
+++ b/src/app/dataset.ts
@@ -115,6 +115,7 @@ export class Dashboard {
     public layout?: string = '';
     public tables?: {[key: string]: string} = {};
     public fields?: {[key: string]: string} = {};
+    public filters?: {[key: string]: any} = {};
     public visualizationTitles?: {[key: string]: string} = {};
     public options?: DashboardOptions = new DashboardOptions();
     public fullTitle?: string; // added to dashboard in validateDashboards()
@@ -122,6 +123,9 @@ export class Dashboard {
     // (sans choices object references) needed to traverse back up Dashboard object
     public relations?: (string | string[])[][];
     public contributors?: {[key: string]: Contributor} = {};
+    // The datastores and layoutObject properties are assigned by the DatasetService.
+    public datastores?: Datastore[] = [];
+    public layoutObject?: any[] = [];
 }
 
 /**
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -77,8 +77,7 @@
 
     <div class="header-toggle-container">
         <div class="dashboard-selector-container" id="dashboard.selector">
-            <app-dashboard-selector (activeDatasetChanged)="simpleFilter.updateSimpleFilterConfig(); toggleDashboardSelectorDialog(false)">
-            </app-dashboard-selector>
+            <app-dashboard-selector [dashboards]="dashboards"></app-dashboard-selector>
             <div class="header-toggle-footer"></div>
         </div>
     </div>
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -76,7 +76,7 @@
     </div>
 
     <div class="header-toggle-container">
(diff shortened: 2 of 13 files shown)


Konnoke (approved):

src/app/dataset.ts
```diff
@@ -115,13 +115,17 @@ export class Dashboard {
     public layout?: string = '';
     public tables?: {[key: string]: string} = {};
     public fields?: {[key: string]: string} = {};
+    public filters?: {[key: string]: any} = {};
```
rartiss55, line 118: Does this mean we are going to have the ability to preset filters through the config?

src/app/dataset.ts
```diff
@@ -115,13 +115,17 @@ export class Dashboard {
...
     public visualizationTitles?: {[key: string]: string} = {};
     public options?: DashboardOptions = new DashboardOptions();
     public fullTitle?: string; // added to dashboard in validateDashboards()
     public pathFromTop?: string[]; // added to dashboard in validateDashboards() - contains keys
     // (sans choices object references) needed to traverse back up Dashboard object
     public relations?: (string | string[])[][];
     public contributors?: {[key: string]: Contributor} = {};
+    // The datastores and layoutObject properties are assigned by the DatasetService.
+    public datastores?: Datastore[] = [];
+    public layoutObject?: any[] = [];
```
rartiss55, line 128: Do we have to add the entire layoutObject in here instead of just using some sort of key to link the dashboard and layout together?

rartiss55:
Not a huge deal, but noticed that for a second now on start up, you see the dashboard selector with no options filled in (don't think that used to happen).

src/app/dataset.ts
ThomasSchellenbergNextCentury: Yep!  :)

It can look something like this:

```
  filters:
  - name: "Uyghur LDC, July 2018 / UI out / Need Type = Food"
    optional: true
    datastore: ""
    database: "ldc_uyg_jul_18"
    table: "ui_out"
    field: "topic"
    operator: "="
    value: "Food"
```

ThomasSchellenbergNextCentury: I realized it might make sense to allow you to use table and field keys with the filters, so I've made a ticket:  https://nextcentury.atlassian.net/browse/THOR-1160

ThomasSchellenbergNextCentury:
@rartiss55 Fixed the issue with seeing the dashboard selector on page load.

rartiss55:
@ThomasSchellenbergNextCentury I guess not showing the selector at all would work if a default is selected in the config (prior to these changes, the dashboard selector autopopulated correctly but now it doesn't -- not sure why).

Any thoughts on layoutObject vs key? I think that's the only comment I had left.

src/app/dataset.ts
ThomasSchellenbergNextCentury: The layoutObject was for a future feature that I had intended to include.  First, it saves a unique copy of each layout associated with each dashboard (since the same config layout can be used by multiple dashboards).  Second, it allows us to save the layout in case the user makes changes like deleting or resizing widgets.  I've just pushed a PR for the changes:  https://github.com/NextCenturyCorporation/neon-dash-internal/pull/207

ThomasSchellenbergNextCentury:
The dashboard selector still auto-populates, it just does so after it got hidden (well, prior to my latest commit).  I think the timing changed from (populate selector, show selector, hide selector) to (show selector, hide selector, populate selector) because previously the dashboard selector populated itself on-init while now it populates itself when it receives a message.

src/app/dataset.ts
rartiss55: Okay, that makes more sense. Thanks for clarifying.

rartiss55 (approved):

PR #196, opened 2019-04-29
Fixing Saved State (THOR Tickets 1021, 1094, 1097, 1122)

ThomasSchellenbergNextCentury:
This PR resolves a few separate tickets:
- 1097:  Using the new neon-server to save and load states to/from flat files.  The default "state directory" is in the root of your neon-server folder (`neon-server/states`).  The saved state files are the same format as Neon Dashboard config files and should be interchangeable.  You can add, delete, or change the local saved state files to affect the Neon Dashboard live.
- 1021:  You can now set dashboard filters in the config file (in the final-level dashboard under "dashboards").  This works both for the saved state files and for the normal Neon Dashboard config file (in `src/app/config/config.*`).
- 1094:  Loaded states are shown nicely in the dashboard selector dropdowns.
- 1122:  Removed the dependency of the Save State Component on the Parameter Service.

You'll need to test this PR using the new Neon Server:  https://github.com/NextCenturyCorporation/neon-server

See the README for setup instructions.  Here's a basic guide on how to run it locally with the Neon Dashboard:
- In `neon-server/blob/master/server/src/main/resources/application.properties`, set `server.port` to an open port (like `8090`)
- In `neon-dash-internal/blob/master/proxy.conf.json`, set `target` to the same port (like `"http://localhost:8090"`)
- Build (`./gradlew build`) and run (`./gradlew bootRun`) the neon-server normally
- Run (`npm start`) the neon-dash-internal normally


ThomasSchellenbergNextCentury:
@s0sterUX Leaving a reminder here for both of us to review the fixed saved state behavior and confirm that it holds up to your expectations.

ThomasSchellenbergNextCentury:
Basic workflow:
1. Load the Neon Dashboard.
2. Do things:  set filters, change widget config, add/delete/move/resize widgets.
3. Click:  hamburger menu (top-right), Saved States.  Then Save a State.
4. You should see your state name in the Load and Delete State dropdowns.  You should also see a YAML file with your state name in your `neon-server/states` folder.
5. Do more things to change the dashboard again (see #2 above).
6. Load your previously-saved state.  The dashboard should return to that state.

src/app/components/save-state/save-state.component.ts
```diff
@@ -286,11 +294,28 @@ export class SaveStateComponent implements OnInit {
         });
     }
 
-    public openNotification(stateName: String, actionName: String) {
-        let message = 'State "' + stateName + '" has been ' + actionName;
+    private openConnection(): any {
+        return this.connectionService.createActiveConnection(this.datasetService.getDatastoreType(),
+            this.datasetService.getDatastoreHost());
```
ScottWarner: This may be a limitation of the Neon library, but if there is no dataStoreType or dataStoreHost connection will be null, and it won't list any saved states

ThomasSchellenbergNextCentury: You are correct.  By funny coincidence, I recently made a ticket for this issue:  https://nextcentury.atlassian.net/browse/THOR-1142

Great minds, eh?

src/app/dataset.ts
```diff
@@ -115,7 +115,7 @@ export class Dashboard {
     public layout?: string = '';
     public tables?: {[key: string]: string} = {};
     public fields?: {[key: string]: string} = {};
-    public filters?: {[key: string]: any} = {};
+    public filters?: any[] = [];
```
arciisine, line 118: Is there a high level type (even nominal) that we can use to help provide some feedback?  E.g. neon.Filter or something to that extent.

src/app/services/dataset.service.ts
```diff
@@ -32,18 +32,20 @@ export class DatasetService {
 
     private static DASHBOARD_CATEGORY_DEFAULT: string = 'Select an option...';
 
-    private datasets: Datastore[] = [];
+    protected datasets: Datastore[] = [];
```
arciisine, line 35: protected doesn't exist in javascript, and won't.  I'd consider public (which is what protected will turn to at runtime) or adding methods to interact.

src/app/services/dataset.service.ts
```diff
@@ -58,6 +60,19 @@ export class DatasetService {
     // ---
     // STATIC METHODS
     // --
+    static appendDashboardChoicesFromConfig(oldChoices: { [key: string]: Dashboard }, newChoices: { [key: string]: Dashboard }): void {
```
arciisine: I'm not fully seeing the append here, this looks more like a deep assignment, and lodash already functionality that does this as well, _.merge.

src/app/services/dataset.service.ts
```diff
@@ -348,8 +363,10 @@ export class DatasetService {
 
         this.datasets = DatasetService.appendDatastoresFromConfig(config.datastores || {}, []);
 
+        this.layouts = config.layouts || {};
+
         DatasetService.updateDatastoresInDashboards(this.dashboards, this.datasets);
-        DatasetService.updateLayoutInDashboards(this.dashboards, config.layouts || {});
+        DatasetService.updateLayoutInDashboards(this.dashboards, this.layouts);
```
arciisine, line 370: We can probably figure out what this needs to be, but with the new config service, it might be better to broadcast a change in config, and let that trickle out globally.

src/app/services/filter.service.ts
```diff
@@ -152,6 +153,43 @@ export namespace FilterUtil {
         return [];
     }
 
+    /**
+     * Creates and returns a filter design from the given JSON object.
+     *
+     * @arg {any} filterObject
+     * @return {FilterDesign}
+     */
+    export function createFilterDesignFromJsonObject(filterObject: any, datasetService: DatasetService): FilterDesign {
```
arciisine, line 162: It looks like the filter object has some presumed structure, we should create an interface to allow for better reference of the assumptions.

src/app/services/filter.service.ts
```diff
@@ -190,6 +228,38 @@ export namespace FilterUtil {
         return filter;
     }
 
+    /**
+     * Creates and returns a JSON object from the given filter design.
+     *
+     * @arg {FilterDesign} filterDesign
+     * @return {any}
+     */
+    export function createFilterJsonObjectFromDesign(filter: FilterDesign): any {
```
arciisine, line 237: Same here, we shouldn't be returning any, it looks like there are 3 possible returns A | B | null, and typescript will let you specify types in the manner (type union)

src/app/services/filter.service.ts
```diff
@@ -731,6 +801,15 @@ export class FilterService {
             this.filterCollection.getFiltersInSingleList(globalDataSource)), [] as AbstractFilter[]).map((filter) => filter.toDesign());
     }
 
+    /**
+     * Returns the filters as JSON objects to save in a config file.
+     *
+     * @return {any[]}
+     */
+    public getFiltersToSaveInConfig(): any[] {
```
arciisine, line 809: This is  returning an array of arrays, e.g., Filter[][], we should be more explicit in what we are returning out

src/app/services/filter.service.ts
```diff
@@ -804,20 +883,31 @@ export class FilterService {
...
+    /**
+     * Sets the filters in the FilterService to the given filter JSON objects from a config file.
      *
-     * @arg {FilterDesign[]} filterDesigns
+     * @arg {any[]} filtersFromConfig
+     * @arg {DatasetService} datasetService
      * @arg {AbstractSearchService} searchService
      */
-    public setFilters(filterDesigns: FilterDesign[], searchService: AbstractSearchService) {
+    public setFiltersFromConfig(filtersFromConfig: any[], datasetService: DatasetService, searchService: AbstractSearchService) {
```
arciisine, line 899: Similar concern about any here

src/app/services/dataset.service.ts
ThomasSchellenbergNextCentury: Well, protected exists in TypeScript:  https://www.typescriptlang.org/docs/handbook/classes.html

Am I misunderstanding your concern here?

src/app/services/dataset.service.ts
ThomasSchellenbergNextCentury: It looks like `_.merge` handles undefined properly differently than we want.  I'll rename the function.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: I could try changing it to `SimpleFilterDesign|CompoundFilterDesign`, but what if it's an incorrectly-formatted object?  I don't want to throw an error, I just want to return null.  I can do the generic `{[key: string]: string}`.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: Good catch.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: 1. I'm pretty sure this is a single array.
2. It's returning a list of JSON objects, not Filters, though I could probably change "any" to "FilterDesign".

src/app/services/dataset.service.ts
ThomasSchellenbergNextCentury: There's probably a lot of logic that can be moved from the DatasetService into the ConfigService.  However, I think it's safer to separate that task into its own ticket.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: Actually, no, that doesn't work, because the `value` can be anything.  So it would have to be `{[key: string]: any}`, which seems (to me) to be the same as `any`.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: Actually, wait:  How would I do the nested filters?  I think the best I can do is this:
`{ name: string, optional: boolean, datastore: string, database: string, table: string, field: string, operator: string, value: any } | { name: string, optional: boolean, type: string, filters: any[] } | null`

src/app/services/dataset.service.ts
arciisine: It does, but not at run time, which is a difference from Java.  Dynamic or untyped code (or cast as any) can violate all protected accesses (and current private, though that should be changing soon).  Ultimately the statement was primarily aimed at the lack of security protected provides, but if being used purely to minimize internal accesses it shouldn't be an issue.

src/app/services/dataset.service.ts
arciisine: You should be able to alter that functionality in _.mergeWith, providing logic to handle custom merging behavior.

src/app/services/filter.service.ts
arciisine: None of the typescript structure affects run-time, only compile time expectations.  And it's possible for the object to not match anything, but then I would ask why are we calling this method with data that is incompatible.  We could look at typeguards to help make assumptions a little clearer.  https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/app/services/filter.service.ts
arciisine: Ah, I misread, and thought the filter function was nested inside of the map operation.  Yeah, its definitely a single array from the outside.

src/app/services/filter.service.ts
arciisine: You can do generic types that can be recursive, let's go over this in person as I think it would go faster.
High level could look like: 
```typescript
type Tree<T> = {
    value: T;
    left: Tree<T>;
    right: Tree<T>;
}
```

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: The data comes from a text file, so it could be literally anything.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: This seems excessive for something that's only used in one place.

ThomasSchellenbergNextCentury:
Folks, I realized that there were a few bugs here with saving moved or resized widgets that I had actually fixed in a separate branch.  I've gone ahead and merged that branch in.  Sorry for the late addition of code.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: I have some ideas; we can talk about it tomorrow.

src/app/dataset.ts
ThomasSchellenbergNextCentury: I'll add something.

src/app/services/dataset.service.ts
ThomasSchellenbergNextCentury: It doesn't look like _.mergeWith works recursively, so we'll still need to have the assignDashboardChoicesFromConfig function to facilitate the recursive call, and I think we'll still have the same code footprint.

src/app/services/filter.service.ts
ThomasSchellenbergNextCentury: I'm going to pull this out into a separate task because this is directly tied to the typing of objects in the dataset.ts file and I think that will be fairly involved.

PR #197, opened 2019-05-03
THOR-861 Fixed visualization data export

ThomasSchellenbergNextCentury:
Must use the current Neon server (the WAR, not the new Java 11 version).

### DIFF
--- a/src/app/components/data-table/data-table.component.ts
+++ b/src/app/components/data-table/data-table.component.ts
@@ -133,7 +133,8 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
             new WidgetFieldOption('heatmapField', 'Heatmap Field', false),
             new WidgetFieldOption('idField', 'ID Field', false),
             new WidgetFieldOption('sortField', 'Sort Field', true),
-            new WidgetFieldArrayOption('filterFields', 'Filter Field(s)', false)
+            new WidgetFieldArrayOption('filterFields', 'Filter Field(s)', false),
+            new WidgetFieldArrayOption('showFields', 'Show Field(s)', false)
         ];
     }
 
@@ -187,6 +188,7 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
                 prettyName: 'No',
                 variable: 'hide'
             }]),
+            // TODO THOR-1135 (Delete this) The allColumnStatus option is deprecated.  Please use showFields now.
             new WidgetSelectOption('allColumnStatus', 'Show Columns on Reload', 'show', [{
                 prettyName: 'Show All',
                 variable: 'show'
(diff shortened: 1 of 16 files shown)


src/app/components/data-table/data-table.component.ts
```diff
@@ -133,7 +133,8 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
             new WidgetFieldOption('heatmapField', 'Heatmap Field', false),
             new WidgetFieldOption('idField', 'ID Field', false),
             new WidgetFieldOption('sortField', 'Sort Field', true),
-            new WidgetFieldArrayOption('filterFields', 'Filter Field(s)', false)
+            new WidgetFieldArrayOption('filterFields', 'Filter Field(s)', false),
+            new WidgetFieldArrayOption('showFields', 'Show Field(s)', false)
```
ThomasSchellenbergNextCentury, line 137: Due to the way I fixed the export (making a single solution that works with all the visualizations), all exported fields must now be represented as WidgetFieldOptions or WidgetFieldArrayOptions.  The one issue this caused was with how we configure the columns that are shown in the Data Table.  We currently use a couple of options to do this:  allColumnStatus, exceptionsToStatus, and fieldsConfig.

I added a "showFields" WidgetFieldArrayOption that now makes the previous options redundant.  I think this change is beneficial because it expects the config option (as an array of string field names) in the same format that's used by WidgetFieldArrayOptions across the dashboard (including the Data Table's own filterFields).  Therefore I've deprecated the old options, to be fully removed at a later time with THOR-1135 (and this will obviously require corresponding changes to LORELEI and AIDA config files).

Konnoke (commented):
Before these changes, in the config under datatable it'll look like:
```
exceptionsToStatus:
                  - date_field_key
                  - original_text_key
```
Is this how it'll look like after the change?:

```
showFields
          date_field
                  - bindingkey: date_field_key
                  - prettyName: Date Created
                  - isRequired: false
                  - enableInMenu: false
```

If not then can you provide an example of how it'll look like in the config?

ThomasSchellenbergNextCentury:
@Konnoke Examples?!  Oh, I guess so...

It's basically just an array of string field names:

```yaml
showFields:
    - createdAt         # You can use actual field names...
    - original_text_key # ...or you can use field keys defined in the "dashboards"
```

The dashboard has existing code that transforms a list of string field names into a list of FieldMetaData objects; we're already using it with the DataTable's filterFields, the MediaViewer's linkFields, and other places.

You bring up a great point, though:  I need to update the config guide and the config files in the repo!

ThomasSchellenbergNextCentury:
@Konnoke I've pushed the config updates to the most recent LORELEI config files and have also updated the config guide.

Konnoke (approved):

PR #199, opened 2019-05-06
Aida 742

gitnordt:
ARL demo config and small cosmetic change.


src/app/dataset.ts
```diff
@@ -101,5 +101,6 @@ export const MediaTypes = {
     html: 'htm',
     pdf: 'pdf',
     audio: 'aud',
-    maskImage: 'mask'
+    maskImage: 'mask',
+    ltf: 'ltf'
```
ThomasSchellenbergNextCentury: @snorpet What is an LTF file?

bdpham: @ThomasSchellenbergNextCentury It's an LDC specific file for annotations.

ThomasSchellenbergNextCentury: OK.  @snorpet Can we just make this a new binding/option on the ThumbnailGrid -- something like `ignoreMediaTypes`?  I feel uncomfortable about hard-coding program-specific media types (when it's not accompanied by specific features, like the mask images with their sliders/overlays).  That way it will be easy to update on other programs, or if LDC changes their file types to `.ldc` files, or something like that, since it will all be in the config file!

gitnordt: I can give that a try.

gitnordt: Done. Hopefully, this change is what you had in mind.

ThomasSchellenbergNextCentury: Yep, looks great!  Thank you!

ThomasSchellenbergNextCentury (approved):

bdpham (approved):

PR #205, opened 2019-05-10
THOR-948 Fix so that an empty filter is no longer required

ThomasSchellenbergNextCentury:
@arciisine I believe this fixes the issue we discussed yesterday with the dashboard generating an error if no filter is defined.  I was able to reproduce the issue (by removing the empty filters you had added) and eliminated the error with this change.  Please confirm whether this works for you too.

### DIFF
--- a/src/app/components/data-table/data-table.component.spec.ts
+++ b/src/app/components/data-table/data-table.component.spec.ts
@@ -37,7 +37,6 @@ import { SearchServiceMock } from '../../../testUtils/MockServices/SearchService
 import { By } from '@angular/platform-browser';
 import * as neon from 'neon-framework';
 import { initializeTestBed } from '../../../testUtils/initializeTestBed';
-import { WhereWrapper } from '../../services/search.service';
 
 describe('Component: DataTable', () => {
     let component: DataTableComponent,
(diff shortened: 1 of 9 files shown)


src/app/components/data-table/data-table.component.spec.ts
```diff
@@ -670,7 +669,10 @@ describe('Component: DataTable', () => {
 
         expect(component.finalizeVisualizationQuery(component.options, {}, [])).toEqual({
             fields: ['*'],
-            filter: new WhereWrapper(neon.query.or())
+            filter: {
```
arciisine: Shouldn't this support an empty object, or being undefined

src/app/components/news-feed/news-feed.component.spec.ts
```diff
@@ -383,7 +382,10 @@ describe('Component: NewsFeed', () => {
 
         expect(component.finalizeVisualizationQuery(component.options, {}, [])).toEqual({
             fields: ['*'],
-            filter: new WhereWrapper(neon.query.or())
+            filter: {
```
arciisine: Same as above

src/app/components/data-table/data-table.component.spec.ts
ThomasSchellenbergNextCentury: Fixed to I think what you would expect here.

arciisine (approved):

PR #206, opened 2019-05-10
THOR-1043 Non-Primitive Options

arciisine:
Everything should be up and running, and app tests have been resolved.


src/app/components/options-list/options-list.component.html
```diff
@@ -35,6 +35,22 @@
...
+    <div mat-fill class="editor" *ngSwitchCase="'NON_PRIMITIVE'">
+        <label>{{ getOption(bindingKey).prettyName }}</label>
+        <ngx-monaco-editor #editor [options]="{
+                theme: 'vs', 
+                language: 'yaml', 
+                automaticLayout: true, 
+                lineNumbers: true,
+                minimap: {
+                    enabled: false
+                }
```
ThomasSchellenbergNextCentury, line 48: Would these options be better saved as a variable in the .ts file?

src/app/widget-option.spec.ts
```diff
@@ -476,3 +478,46 @@ describe('WidgetOptionCollection with bindings and custom fields', () => {
...
+    it('Objects update properly', () => {
+        const option = new WidgetNonPrimitiveOption('test', 'Test', {}, true);
+        expect(option.valueDefault).toEqual({});
+        option.intermediateValue = 'a: [1,2,3]';
+        expect(option.getValueToSaveInBindings()).toEqual({ a: [1, 2, 3] });
+        option.intermediateValue = `
+a: 5
+b: 10
+c: [1,2,3]
+`;
```
ThomasSchellenbergNextCentury: The indentation here annoys me, but I understand why it is necessary.  Is there a commonly-accepted standard for dealing with situations like this?  Would it be "cleaner" to use `\n` instead?

src/app/widget-option.spec.ts
```diff
@@ -476,3 +478,46 @@ describe('WidgetOptionCollection with bindings and custom fields', () => {
         expect(options.testCustomFieldArray).toEqual([DatasetServiceMock.NAME_FIELD, DatasetServiceMock.TYPE_FIELD]);
     }));
 });
+
+describe('NonPrimitive Fields', () => {
```
ThomasSchellenbergNextCentury, line 482: I think you should add a test on setting `intermediateValue` to an improperly-formatted string and checking that it returns `undefined`.

src/app/widget-option.ts
```diff
@@ -193,6 +194,9 @@ export class WidgetMultipleSelectOption extends WidgetOption {
 }
 
 export class WidgetNonPrimitiveOption extends WidgetOption {
+
+    private _intermediateValue: string;
```
ThomasSchellenbergNextCentury, line 198: I like the `_intermediateValue` here.  I've added a note that we could likely use this strategy on validating numbers in the future too:  https://nextcentury.atlassian.net/browse/THOR-1044

ThomasSchellenbergNextCentury (changes requested):
Please see specific comments for details.

src/app/components/options-list/options-list.component.html
arciisine: Potentially, they are intimately tied to the component.  I view them akin to html attributes, vs external config.

src/app/widget-option.spec.ts
arciisine: Yaml will consume all the leading spaces of equal width, it just may not be apparent that is happening when comparing the values.  We could also build a function to strip the leading spaces consistently.

src/app/widget-option.spec.ts
arciisine: Sure, will add.

src/app/widget-option.spec.ts
arciisine: Relied on yaml stripping indentation

ThomasSchellenbergNextCentury (approved):

rartiss55 (approved):

Konnoke (approved):

PR #207, opened 2019-05-14
THOR-1162 Save Widget Layout Changes

ThomasSchellenbergNextCentury:
If the user changes the dashboard layout (delete/move/resize), then selects a different dashboard, then returns to the first dashboard, they now return to the modified layout.


ThomasSchellenbergNextCentury:
@rartiss55 I'm merging this because something seems to be broken with the saved state behavior on THOR-1021 without the updates here.

rartiss55:
@ThomasSchellenbergNextCentury that's fine, just looked over it. Consider this comment retroactive approval :)

PR #214, opened 2019-05-20
AIDA-800 ARL Sub-Subtypes for Taxonomy

gitnordt:
Fixed error for subType children undefined, added levels to taxonomy groups in order to perform an exact match check and get accurate counts for subTypes, fixed issue with parent siblings not getting added to filters, code cleanup


src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -660,17 +685,19 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
     retrieveUnselectedNodes(nodeArray: any[]) {
         let relatives = [];
         for (let node of nodeArray) {
-            //Ensures that only node child relatives(with checkboxes) are added and not the values listed(without checkboxes)
-            if (node.children && node.children.length && node.children[0].description !== this.options.valueField.columnName) {
+            //Ensures that only child nodes(with checkboxes) are added and not leaf nodes(without checkboxes)
+            if (node.children && node.children.length) {
                 for (let child of node.children) {
-                    if (child.checked === false && child.description !== this.options.valueField.columnName) {
+                    if ((child.checked === false || child.indeterminate === true)
```
ThomasSchellenbergNextCentury: I know this is nitpicking, but is there a reason we're using `child.checked === false` and `child.indeterminate === true` rather than `!child.checked` and `child.indeterminate`?

gitnordt: Definitely nitpicking being that this is valid code and preferred in some cases LOL. Fortunately not by me, I prefer shorthand and usually do use it. The only reason that I can think of why I did it this way was because I really, really felt like typing the code out that day. I'll change it just for you!

ThomasSchellenbergNextCentury (approved):

PR #217, opened 2019-05-21
Merge LORELEI into AIDA / VERDI

ThomasSchellenbergNextCentury:
Hey folks, could you please make sure that the AIDA dashboards / datasets work with the LORELEI team's changes?

As a reminder, you can use the same ES 2 and Neon Server WAR that you've been using.

Overall, the changes include the config overhaul (with the new config files), the new widget option ("gear") menus, the filtering changes, and a variety of other minor enhancements and bug fixes.

I've verified that all of the AIDA-specific filtering works (I think!), but I encourage you to double-check me :)

If you'd prefer, we can wait to merge this until the new filter badges have been implemented (which should be ready within the next few days).

Thanks!


src/app/components/taxonomy-viewer/taxonomy-viewer.component.html
```diff
@@ -168,4 +53,11 @@
         </div>
     </div>
```
gitnordt, line 55: The taxonomy leaf nodes should not be displayed with checkboxes since they are not intended to be filterable. It looks like you removed the columnName from the description in the taxonomy node objects. This has caused the check to for displaying checkboxes to become invalid. Please change node.data.description to be node.data.description.columnName in the html.

gitnordt (changes requested):

src/app/components/document-viewer/document-viewer.component.html
```diff
@@ -176,7 +52,20 @@
...
+    <!-- need to account for footer only, contribution text only, or footer and contribution text together -->
+    <div class="footer" *ngIf="showPagination() && !showContribution()">
+        <div class="footer-button-container">
+            <button mat-raised-button [disabled]="page == 1" class="pagination-button" color="primary" (click)="goToPreviousPage()">
+                Previous
+            </button>
+            <span class="pagination-button-spacer"></span>
+            <button mat-raised-button [disabled]="lastPage" class="pagination-button" color="primary" (click)="goToNextPage()">
+                Next
+            </button>
```
gitnordt, line 64: This may be a separate ticket, but I noticed this issue while testing. The thumbnail grid paging and filtering are not working together properly. When you click on a document on the first page of the thumbnail grid, the filter is created and the document that was clicked is displayed. If you go to another page and perform the same action, the filtered document is not displayed. 

I think it's because that document is still showing up on the first page when we are on the next page if that makes any sense. I remember fixing this issue a while back when Charlotte was interning ([https://nextcentury.atlassian.net/browse/AIDA-278](AIDA-278)), but the fix was removed most likely during a merge.

gitnordt (changes requested):

gitnordt:
I noticed that the legend in the map does not have a switch to turn legendFiltering off in the map like in the network graph. This would be a great addition. Can we put in a ticket to add that option to the map legend? 

Also, I agree that it would be a good idea to add the new badges to this PR.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.html
ThomasSchellenbergNextCentury: Will do!

ThomasSchellenbergNextCentury:
@snorpet "This would be a great addition" I think you meant to say that it would be a LEGENDARY addition!  I created the ticket:  https://nextcentury.atlassian.net/browse/NEON-43

src/app/components/document-viewer/document-viewer.component.html
ThomasSchellenbergNextCentury: Done!

src/app/components/timeline/timeline.component.ts
```diff
@@ -158,152 +180,68 @@ export class TimelineComponent extends BaseNeonComponent implements OnInit, OnDe
...
-                clauses.push(neon.query.where(filterField, '=', value));
-            }
-        }
+        this.selected = [beginDate, endDate];
 
-        return neon.query.or.apply(neon.query, clauses);
-    }
+        if (this.options.filterField.columnName) {
+            let filterValues: any[] = neonUtilities.flatten((selectedData || []).map((selectedItem) => selectedItem.filters)).filter(
+                (value, index, array) => array.indexOf(value) === index);
```
gitnordt, line 231: The new filterField code does not seem to be working. On the TA2 Network Dashboard, when I select a document in the right-side component, I expect the document component as well as the timeline to filter. The timeline is no longer filtering when a document is clicked.

gitnordt (changes requested):

src/app/components/document-viewer/document-viewer.component.html
gitnordt: The document is now showing up as expected, but when the filter is removed, I am taken to the first page again. Please add the code back in to keep track of the current page. It is not ideal for the user to click through 5 pages, used the filter, and then have to click through the pages again and try to find where they left off.

src/app/components/timeline/timeline.component.ts
gitnordt: Wait, I'm wrong. I don't believe that it is the filterField causing this issue. Document/Timeline filtering works as expected in TA2 Events and Map Dashboard. We may need to dig a little bit to find out why this isn't working.

src/app/components/document-viewer/document-viewer.component.html
ThomasSchellenbergNextCentury: Stacey, I don't take your code away from you, I just break your code!  :)

Anyway, I can totally understand the need for this behavior.  It should be fixed now.  Sorry for breaking it in the first place!

src/app/components/timeline/timeline.component.ts
ThomasSchellenbergNextCentury: I saw it too.  I think I've fixed the issue.

gitnordt (approved):

bdpham (approved):

PR #221, opened 2019-05-24
Thor taxonomy optimization

arciisine:
This is the initial pass at the taxonomy overhaul, and was hoping to see if you could kick the tires a bit and let me know if anything seems off.  The counts are still a little different so I'm assuming there is still some stuff to discuss.


src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -298,203 +408,53 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
-                        }
-
-                        this.sortTaxonomyArrays(this.taxonomyGroups[foundCategory.index].children[foundType.index].children);
-                    }//end types loop
+            const child = {
+                description: this.options.valueField,
+                name: find(d, 'valueField') || find(d, 'idField'),
+                sourceIds: find(d, 'sourceIdField'),
+                externalId: find(d, 'idField')
+            };
```
gitnordt, line 416: Your valueField logic does not appear to be working. When this field is set, you should see a final leaf layer that only consists of values. Its styling is completely different. The values are surrounded by a border and do not have checkboxes because they are listed for discovery purposes and not intended to be filterable. 
valuesScreenshot

src/app/components/taxonomy-viewer/taxonomy-viewer.component.html
```diff
@@ -21,36 +21,58 @@
     </div>
 
     <div class="body-container" [class.with-contribution]="showContribution()">
-        <div #taxonomyViewer class="taxonomy-viewer-container">
+        <mat-tree [dataSource]="dataSource" [treeControl]="treeControl">
+            <mat-tree-node *matTreeNodeDef="let node" matTreeNodeToggle>
```
gitnordt: In all honesty, I definitely like the styling of the Angular Tree better than the Material Tree. The checkboxes are a little too vibrant for my taste. I also do understand the need to minimize the Neon code base. I think ultimately this is Sophie and Thomas' call.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.html
```diff
@@ -21,36 +21,58 @@
...
                 <ng-template #treeNodeFullTemplate let-node let-index="index" let-templates="templates">
                     <div [ngClass]="setClassForTreePosition(node, 'tree-node tree-node-level-')" [class]="node.getClass()">
                         <div class="node-wrapper" [style.padding-left]="node.getNodePadding()">
                             <tree-node-expander [node]="node"></tree-node-expander>
                             <div class="node-content-wrapper" (click)="node.mouseAction('click', $event)"
-                                 (dblclick)="node.mouseAction('dblClick', $event)"
-                                 [class.node-content-wrapper-active]="node.isActive"
-                                 [class.node-content-wrapper-focused]="node.isFocused">
+                                (dblclick)="node.mouseAction('dblClick', $event)" [class.node-content-wrapper-active]="node.isActive"
+                                [class.node-content-wrapper-focused]="node.isFocused">
```
gitnordt: It looks like the double click feature that used to open and close the tree levels is being overridden, I'm assuming by the toggle filtering action that Thomas put in. It was a nice to have, but not a necessity so feel free to remove the dblclick code since it is no longer being used.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -288,6 +314,90 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
+        if (!(child.externalId in currentGroup.childrenMap)) {
+            currentGroup.childrenMap[child.externalId] = child;
+            // currentGroup.children.push(child);
+            child.parent = currentGroup;
+            child.level = pos + 1;
+
+            // Walk back up if a new item
+            while (currentGroup && currentGroup.id) {
+                currentGroup.count += 1;
+                if (child.externalId) {
```
gitnordt: The counts appear to be inaccurate because duplicate nodeIds(kbids) are being counted. Unfortunately, the nodeIds are listed multiple times in the database to represent each link to the node. You may want to add the check back in to track nodeIds and ensure that they are not duplicated/counted multiple times. You may also need a check for the name and lineage to ensure you are counting the correct child in the tree. You can reference line 482 of the previous code. Or in hindsight, I guess you could also use the ElasticSearch _id instead since it is unique, but you would still need to collect the nodeIds for filtering purposes.

gitnordt (changes requested):
Yaaay! The taxonomy load performance is much improved. I just had a few suggestions to correct some of the functionality/features and a couple comments.

arciisine:
Closing in favor of https://github.com/NextCenturyCorporation/neon-dash-internal/pull/224, code will remain, but PR is dead for the time being

PR #222, opened 2019-05-29
THOR - Data Set Notifications With Counts

arciisine:


arciisine:
Ties into https://github.com/NextCenturyCorporation/neon-server/pull/11

src/app/dashboard/abbreviate.pipe.ts
```diff
@@ -0,0 +1,22 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+const suffixes = ['k', 'M', 'G', 'T', 'P', 'E'];
+
+@Pipe({
+  name: 'abbreviate'
+})
+export class AbbreviatePipe implements PipeTransform {
```
ThomasSchellenbergNextCentury, line 8: Should we move this pipe class to a separate folder (like, maybe `src/app/pipes`)?  It seems general-use.

src/app/dashboard/dashboard.component.ts
```diff
@@ -589,8 +589,8 @@ export class DashboardComponent implements AfterViewInit, OnInit, OnDestroy {
     /**
      * Indicates to the dashboard that there is new data available
      */
-    dataAvailableDashboard() {
-        this.updatedData = true;
+    dataAvailableDashboard(event: { message: MessageEvent }) {
+        this.updatedData += (JSON.parse(event.message.data).count || 1);
```
ThomasSchellenbergNextCentury, line 592: @arciisine I think maybe we shouldn't assume that the count is 1 if it's not provided... Could we just show a blank red dot instead?

src/app/dashboard/dashboard.component.html
```diff
@@ -32,9 +32,9 @@
     <mat-toolbar class="neon-toolbar-large neon-toolbar-wide titlebar-header-text" color="primary" layout-align="space-between center" flex>
         <span>{{ datasetService.getCurrentDashboardTitle() || "Choose Dashboard..." }}</span>
         <span class="fill-remaining"></span>
-        <button mat-icon-button class="neon-icon-button-large notify" aria-label="Notifications" matBadgePosition="above after"
-            [matBadgeHidden]="!updatedData" matBadgeColor="warn" matBadge="x" matBadgeSize="small" (click)="refreshDashboard()"
-            matTooltip="Refresh Data" tooltip-position="below">
+        <button mat-icon-button class="neon-icon-button-large notify" aria-label="Notifications" matBadgePosition="above before"
```
ThomasSchellenbergNextCentury, line 35: @arciisine I'm getting "Maximum call stack size exceeded" errors on this line when I press the bell button.

arciisine: I'll take a look, not sure what would be recursive there

src/app/dashboard/abbreviate.pipe.ts
arciisine: I generally would put it closest to it's immediate use, and pull into a more common place when the use case arises.  Specifically to limit the amount of extra general code we are putting in the common area that may only be used by one component.

src/app/dashboard/dashboard.component.ts
arciisine: The badge doesn't like it when the text is missing.  Additionally, if I get a mixture of countable and non-countable events, how do I notify the user of changes.  I'd rather go the other way and make count a required field on notification.

src/app/dashboard/dashboard.component.html
arciisine: The button was triggering a refresh event, and refresh events were being listened to by the dashboard, and calling the same method (hence recursion).  We should be good now.

ThomasSchellenbergNextCentury (approved):

PR #224, opened 2019-05-30
THOR-1223 Taxonomy optimization, v2

arciisine:
Back ported changes from https://github.com/NextCenturyCorporation/neon-dash-internal/pull/221 to mitigate need for test overhaul. Should be near complete, needs final verification on newest subtype support.


src/app/components/taxonomy-viewer/taxonomy-viewer.component.spec.ts
```diff
@@ -33,7 +33,7 @@ import { TaxonomyViewerComponent } from './taxonomy-viewer.component';
 import { TaxonomyViewerModule } from './taxonomy-viewer.module';
 import { ConfigService } from '../../services/config.service';
 
-describe('Component: TaxonomyViewer', () => {
+fdescribe('Component: TaxonomyViewer', () => {
```
bdpham: Looks like you forgot to remove the fdescribe.

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -284,166 +308,324 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
+                currentGroup = currentGroup.parent;
+            }
+        }
+    }
+
+    /**
+     * Navigate each level, sorting by name if children present
+     */
+    sortTaxonomies(group: TaxonomyGroup | TaxonomyNode) {
+        if ('children' in group) {
```
ThomasSchellenbergNextCentury, line 393: Out of curiosity, why is this method preferable to checking if `group.children` is defined?

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -284,166 +308,324 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
+                description: this.options.valueField,
+                name,
+                sourceIds: find(data, 'sourceIdField'),
+                externalId: find(data, 'idField'),
+                externalName: name
+            };
+
+            // Loop, categories[] -> types[] -> subTypes?[]
+            for (const category of categories) {
+                for (const type of types) {
```
ThomasSchellenbergNextCentury, line 456: If not now, we may want to add a ticket about re-implementing sub-type field support in the future, if/when it ever becomes relevant.

arciisine: I think this is the outstanding item from Stacey's PR that we would want to resolve.  The merge function already handles it, so it should theoretically be straightforward

src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
arciisine: In typescript, the in check can actually narrow types, by testing for whether or not a property is defined on an object (vs it's value), https://github.com/Microsoft/TypeScript/issues/10485

ThomasSchellenbergNextCentury:
@arciisine I don't think the subtype filters are working correctly.  For example, if I uncheck `RELATION (202)`, then I check `OPRA (5)` (under `RELATION (202) -> General Affiliation (71)`), I should only see 5 yellow boxes, but I actually see a lot more than that.  I think it's showing me all 71 `General Affiliation` relations, even though most of them should be filtered-out.  Can you please investigate?

gitnordt:
> @arciisine I don't think the subtype filters are working correctly. For example, if I uncheck `RELATION (202)`, then I check `OPRA (5)` (under `RELATION (202) -> General Affiliation (71)`), I should only see 5 yellow boxes, but I actually see a lot more than that. I think it's showing me all 71 `General Affiliation` relations, even though most of them should be filtered-out. Can you please investigate?

Same here, the only filtering that appears to be working properly is the category filtering at the parent level. The subTypes and multi-level subTypes are not filtering as expected.

arciisine:
Porting the alternate version back, a small piece was missed, which has now been resolved.  Because how the names get broken down, the full dotted name must be preserved, and is under the 'externalName' field.  This should have been passed to the filter but was not. @snorpet

ThomasSchellenbergNextCentury (dismissed):

gitnordt (dismissed):
The errors I found were addressed and tests are passing.

PR #226, opened 2019-05-30
THOR-1232 New News Feed UI

arciisine:
All feedback should be resolved, aida interoperability should be supported, but will be a bigger discussion..


src/app/components/base-neon-component/base-neon.component.ts
```diff
@@ -1160,9 +1160,9 @@ export abstract class BaseNeonComponent implements AfterViewInit, OnInit, OnDest
             bindings[option.bindingKey] = option.getValueToSaveInBindings();
             return bindings;
         }, {
-            layers: (options || this.options).layers.length ? (options || this.options).layers.map((layer) => this.getBindings(layer)) :
-                undefined
-        });
+                layers: (options || this.options).layers.length ? (options || this.options).layers.map((layer) => this.getBindings(layer)) :
```
ThomasSchellenbergNextCentury, line 1163: This is weird.  Why is your editor even adding indentation here?

src/app/components/news-feed/news-feed.component.ts
```diff
@@ -93,16 +94,23 @@ export class NewsFeedComponent extends BaseNeonComponent implements OnInit, OnDe
...
         return [
             new WidgetFieldOption('contentField', 'Content Field', false),
+            new WidgetFieldOption('sourceContentField', 'Source Content Field', false),
+            new WidgetFieldOption('titleContentField', 'Title Content Field', false),
             new WidgetFieldOption('dateField', 'Date Field', false),
             new WidgetFieldOption('filterField', 'Filter Field', false),
             new WidgetFieldOption('idField', 'ID Field', true),
             new WidgetFieldOption('linkField', 'Link Field', false),
-            new WidgetFieldOption('primaryTitleField', 'Primary Title Field', false),
-            new WidgetFieldOption('secondaryTitleField', 'Secondary Title Field', false),
```
ThomasSchellenbergNextCentury, line 101: You've got to be careful about renaming old config options:  now the AIDA team has to edit all of their config files.

It's usually preferable to add behavior to support backwards compatibility just in case they still have primaryTitleField and/or secondaryTitleField configured.

src/app/components/news-feed/news-feed.component.ts
```diff
@@ -93,16 +94,23 @@ export class NewsFeedComponent extends BaseNeonComponent implements OnInit, OnDe
     createFieldOptions(): (WidgetFieldOption | WidgetFieldArrayOption)[] {
         return [
             new WidgetFieldOption('contentField', 'Content Field', false),
+            new WidgetFieldOption('sourceContentField', 'Source Content Field', false),
```
ThomasSchellenbergNextCentury: The name "source content" is a bit ambiguous.  I think your intent is that it's the original text, but we shouldn't name a config option like that, since that's very domain-specific.

src/app/components/news-feed/news-feed.component.html
```diff
@@ -20,27 +19,35 @@
...
-                        {{ item[options.contentField.columnName] }}</div>
-                    <div class="content-footer">
-                        <p class="text-highlight">{{ item[options.dateField.columnName] }}</p>
-                        <p [hidden]="!item[options.secondaryTitleField.columnName]">{{ item[options.secondaryTitleField.columnName] }}</p>
+        <mat-accordion class="news-feed-container" *ngIf="newsFeedData != null && newsFeedData.length" displayMode="flat">
+            <mat-expansion-panel class="news-feed-div" *ngFor="let item of newsFeedData" (click)="selectGridItem(item)">
+                <mat-expansion-panel-header [collapsedHeight]="'35px'" [expandedHeight]="'35px'">
+                    <mat-panel-title>
+                        <span class="date">{{ relativeTime(item[options.dateField.columnName]) }}</span>
+                        <span class="text">{{ item[options.titleContentField.columnName] }}</span>
```
ThomasSchellenbergNextCentury: It looks like we're assuming that specific fields are always present, like dateField and titleContentField.  But what if those fields are not configured?  We should be able to handle those cases gracefully.  Plus, we should verify in the validateVisualizationQuery function that all required fields have been configured (currently, it's just the idField).

src/app/components/news-feed/news-feed.component.html
```diff
@@ -20,27 +19,35 @@
...
+                    </mat-panel-title>
+                </mat-expansion-panel-header>
+                <mat-panel-description>
+                    <div>
+                        <dl *ngIf="item[options.contentField.columnName]">
+                            <dt>Text</dt>
```
ThomasSchellenbergNextCentury: We shouldn't be hard-coding field names like this.  They should either be taken from the FieldMetaData object (for example, here it would be options.contentField.prettyName) or be defined in a config option.

src/app/components/news-feed/news-feed.component.html
```diff
@@ -20,27 +19,35 @@
...
+                            <dd>{{item[options.contentField.columnName]}}</dd>
+                        </dl>
+                        <dl *ngIf="item[options.sourceContentField.columnName]">
+                            <dt>Original</dt>
+                            <dd>{{item[options.sourceContentField.columnName]}}</dd>
+                        </dl>
```
ThomasSchellenbergNextCentury: I think the JIRA ticket requested that we add an array of configurable dropdown fields, but it looks like we're only using two specific fields here.  While this may work fine for now, we don't want to have to update the code if we decide we want to add another field to the dropdown in the future, or another team (AIDA/KAIROS) wants to use it.  Grep the repo for WidgetFieldArrayOption to see how this is done in other widgets.

src/app/components/news-feed/news-feed.component.html
arciisine: Yeah, this was due to it being a WIP and wanting to vet design before building it fully out.

src/app/components/news-feed/news-feed.component.html
arciisine: For now the widget supports primary and secondary text fields with configurable labels.

src/app/components/news-feed/news-feed.component.html
arciisine: Okay, and for the news feed I think we should be able to rely upon date field and title field being set, as those are key parts of what makes it news entries.

src/app/components/news-feed/news-feed.component.ts
arciisine: Yeah, in conversation with Stacey/Sophie to figure out what is needed.

src/app/components/base-neon-component/base-neon.component.ts
arciisine: Dunno, using the eslinting rules, but things still seem slightly off.

arciisine:
Below is the config I used for testing the news feed 

```yaml
- name: News Feed
          type: newsFeed
          icon: ViewData64
          col: 1
          row: 1
          sizex: 4
          sizey: 14
          bindings:
              title: Document Feed
              tableKey: table_key_1
              idField: id_field_key
              sortField: date_field_key
              filterField: date_field_key
              limit: 100
              dateField: date_field_key
              contentField: translated_text_key
              secondaryContentField: original_text_key
              titleContentField: translated_text_key
              secondaryContentLabel: Original Text
            #   filterable: false
            #   filterFields:
            #       - filter_field_key
              singleFilter: true
              ignoreSelf: false
          minPixelx: 320
          minPixely: 240
          minSizex: 7
          minSizey: 3
```

ThomasSchellenbergNextCentury (approved):

PR #236, opened 2019-06-05
THOR-1023 - Step 3 - Standardize Neon Config Usage

arciisine:
Unified config around single data model, dealing with discrepancies and resolving logical issues.


src/app/components/data-table/data-table.component.ts
```diff
@@ -342,7 +342,7 @@ export class DataTableComponent extends BaseNeonComponent implements OnInit, OnD
                     // If showFields is populated, hide each field that is not in showFields (override allColumnStatus).
                     active: this.options.showFields.length ? false : (this.options.allColumnStatus === 'show'),
                     style: {},
-                    width: this.getColumnWidth(fieldObject.name)
+                    width: this.getColumnWidth(fieldObject['name']) // TODO: Investigate
```
ThomasSchellenbergNextCentury, line 345: @arciisine This should probably be fieldObject.columnName

src/app/components/save-state/save-state.component.spec.ts
```diff
@@ -162,36 +162,41 @@ describe('Component: SaveStateComponent', () => {
...
             fileName: 'dashboard1',
+            version: '1',
             lastModified: Date.now(),
             dashboards: {
                 name: 'dashboard1'
             },
             datastores: {
                 datastore1: {}
-            } as any,
+            } as any as Record<string, NeonDatastoreConfig>, // TODO: Verify typings
```
ThomasSchellenbergNextCentury, line 172: This should be an object read from a saved state file (so the same format as the config.yaml/config.json file)

src/app/neon-grid-item.ts
```diff
@@ -26,6 +26,11 @@ export interface NeonGridItem extends NeonLayoutGridConfig {
     type?: string;
     name?: string;
```
ThomasSchellenbergNextCentury, line 27: @arciisine  FYI you should be able to remove `hide`, `id`, `icon`, `type`, and `name` (unless any of them are injected by Angular2-Grid, I guess).

src/app/services/dashboard.service.spec.ts
```diff
@@ -100,14 +98,14 @@ describe('Service: DashboardService Static Functions', () => {
...
-        let dashboard = new Dashboard();
+        let dashboard = Dashboard.get();
         dashboard.name = 'name';
         DashboardService.assignDashboardChoicesFromConfig(input, {
             test1: {
                 choices: {
                     test2: dashboard
                 }
-            }
+            } as any as Dashboard // TODO: Typings verify
```
ThomasSchellenbergNextCentury, line 108: For `assignDashboardChoicesFromConfig`, I believe `oldChoices` are an internal-state-dashboard and `newChoices` are a dashboard-from-config-file.

src/app/types.ts
```diff
@@ -15,7 +17,7 @@
  */
 // TODO: THOR-825: rename classes/functions that still reference 'dataset' to say 'datastore' (THOR-1052)
 
-export class FieldMetaData {
+export class FieldMetaData implements NeonFieldMetaData {
```
ThomasSchellenbergNextCentury, line 20: @arciisine Some of the classes here like FieldMetaData, TableMetaData, and DatabaseMetaData:  should they be completely replaced with their interface counterparts?  I guess the one concern there is setting default values, but that could be done using a static builder method... but maybe that's a separate ticket.

src/app/neon-gtd-config.ts
```diff
@@ -89,37 +72,48 @@ export interface NeonDashboardConfig {
...
+    name?: string;
     host: string;
     type: string;
     databases: Record<string, NeonDatabaseMetaData>;
 }
 
-export interface NeonGTDConfig {
-    projectTitle: string;
-    projectIcon: string;
+export interface NeonGTDConfig<T extends NeonDashboardConfig<any> = NeonDashboardConfig<any>> {
```
ThomasSchellenbergNextCentury, line 106: @arciisine For NeonGTDConfig and Dashboard, you have both an interface and a class with the same name... How does that work, and why is that useful?

src/app/neon-gtd-config.ts
```diff
@@ -89,37 +72,48 @@ export interface NeonDashboardConfig {
...
 }
 
 export interface NeonDatastoreConfig {
+    name?: string;
     host: string;
     type: string;
     databases: Record<string, NeonDatabaseMetaData>;
 }
 
-export interface NeonGTDConfig {
```
ThomasSchellenbergNextCentury, line 115: BTW feel free to rename NeonGTDConfig.  The original Neon UI was called the "Geo-Temporal Dashboard" and the acronym stuck even though it's not completely accurate any more.  Perhaps "NeonDashboardConfigFile" would be better now?

src/app/neon-gtd-config.ts
```diff
@@ -89,37 +72,48 @@ export interface NeonDashboardConfig {
...
     bindings?: Record<string, any>;
     sizex: number;
     sizey: number;
-    minPixelx?: number;
-    minPixely?: number;
-    minSizex?: number;
-    minSizey?: number;
 }
 
 export interface NeonLayoutConfig extends NeonLayoutGridConfig {
```
ThomasSchellenbergNextCentury, line 94: @arciisine So what's the difference between NeonLayoutGridConfig, NeonLayoutConfig, and NeonGridItem?  Is it:
- NeonLayoutGridConfig: Properties from the Neon config file needed by Angular2-Grid
- NeonLayoutConfig: Properties in the Neon config file needed by Neon
- NeonGridItem: Properties NOT in the Neon config file needed by Angular2-Grid

If so, you can move "bindings" to NeonLayoutConfig since that's not an Angular2-Grid specific property.

src/app/neon-gtd-config.ts
```diff
@@ -89,37 +72,48 @@ export interface NeonDashboardConfig {
...
-    name: string;
+    name?: string;
     type: string;
 }
 
 export interface NeonDatastoreConfig {
+    name?: string;
```
ThomasSchellenbergNextCentury, line 100: I don't think the Datastore needs a name.

src/app/neon-gtd-config.ts
```diff
@@ -89,37 +72,48 @@ export interface NeonDashboardConfig {
...
+
     datastores: Record<string, NeonDatastoreConfig>;
-    dashboards: NeonDashboardConfig;
-    layouts: Record<string, NeonLayoutConfig>;
-    errors: string[];
-    neonServerUrl: string;
+    dashboards: T;
+    layouts: Record<string, NeonLayoutConfig[]>;
+    errors?: string[];
+    neonServerUrl?: string;
```
ThomasSchellenbergNextCentury, line 116: @arciisine The NeonGTDConfig represents the config file, right?  In that case, "visualizationTitles" and "stateName" should be here, and I don't think "errors" are needed/used.

src/app/types.ts
arciisine: Separated out in the next PR to allow for config to serialize easily.

src/app/neon-gtd-config.ts
arciisine: This creates a type space that allows for attachment of static methods, but without a constructor enforces usage of interfaces only.

src/app/neon-gtd-config.ts
arciisine: Renamed to NeonConfig, as we already had NeonDashboardConfig referring specifically to a dashboard structure.

src/app/neon-gtd-config.ts
arciisine: Cool, done.

src/app/neon-gtd-config.ts
arciisine: That was relied upon when the data stores were in array form, the key of the map became the name.  We can probably discuss a little further.

src/app/neon-gtd-config.ts
arciisine: What specifically is stateName, and visualizationTitles are currently in the dashboard config, should those be moved out?

src/app/neon-gtd-config.ts
ThomasSchellenbergNextCentury: Oh, interesting.  Thanks for the explanation!

src/app/neon-gtd-config.ts
ThomasSchellenbergNextCentury: Sounds good.

src/app/neon-gtd-config.ts
ThomasSchellenbergNextCentury: Ah, OK, that makes sense.  Forget my comment then.

src/app/neon-gtd-config.ts
ThomasSchellenbergNextCentury: Re: stateName, it's added by the SaveStateComponent to the config object that's sent to the Neon Server, and I think we've determine that the config object should be a NeonGTDConfig, since it's being saved-to-file?

Re: visualizationTitles, they might just need to be copied?  As long as we can read them from a file and use them in the widgets like we're doing now.  Here's an example:  https://github.com/NextCenturyCorporation/neon-dash-internal/blob/master/src/app/config/config.darpa-July2018-es6.yaml#L187

src/app/services/dashboard.service.spec.ts
arciisine: Have resolved this with the .get method and deeppartials

PR #237, opened 2019-06-05
THOR-1140 Updating Copyrights

NBrockman:
Header added to .eslintrc.yaml file to add new copyrights header when eslint is run.

### DIFF
--- a/src/app/components/text-cloud/text-cloud-namespace.ts
+++ b/src/app/components/text-cloud/text-cloud-namespace.ts
@@ -1,5 +1,5 @@
-/*
- * Copyright 2017 Next Century Corporation
+/**
+ * Copyright 2019 Next Century Corporation
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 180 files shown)


src/app/components/text-cloud/text-cloud-namespace.ts
```diff
@@ -11,11 +11,6 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
- *
- */
-
-/*
- * Reimplemented from: jquery.tagcloud.js at https://github.com/addywaddy/jquery.tagcloud.js by Adam Groves Copyright 2008
```
ThomasSchellenbergNextCentury, line 18: @NBrockman This comment looks important; we should probably keep it in here.  Maybe just use a line comment (`//`) instead of a block comment (`/* */`).

NBrockman: Good catch, I'll change that right away.

ThomasSchellenbergNextCentury (approved):

PR #242, opened 2019-06-06
THOR-916 Snackbar Errors

NBrockman:
Added snackbar error messages to handleDashboardError function.


ThomasSchellenbergNextCentury:
@NBrockman I know that Sophie is still thinking about exactly what the errors should say, but I think that falls within the scope of a separate ticket, so I'm going to finalize this ticket now.

ThomasSchellenbergNextCentury (approved):

PR #245, opened 2019-06-07
THOR-1028-Library-Icons-Update

NBrockman:
Updated the icons in assets with new images. Updated the Visualization constants. Moved the old, no longer used icons into a new folder called "old_icons".


ThomasSchellenbergNextCentury:
@NBrockman Overall, this looks good.  A couple of things:
- Only the visualization icons were updated, but some other icons were accidentally deleted.  We need to keep any icons in `src/assets/icons` that don't have new versions.  This includes:  `database_icon`, `filters`, `filters_active`, `neon_logo`, `no_data`, and `view_data` (In the future, we probably want to separate the visualization icons from the other icons, but let's worry about that in a separate ticket).
- We don't need to keep the old icons.  Delete the `old_icons` folder as well as the `__MACOSX` folder.  We can always go back in the git history and find the old icons if we really need them.

NBrockman:
I deleted the old visualization icons. The view_data though was just used as the data-table icon so that was a duplicate so I also deleted view_data as well.

ThomasSchellenbergNextCentury (approved):

PR #247, opened 2019-06-11
THOR-1036-Component-Header-Padding

NBrockman:
Changed all components, in the dashboard grid, to have 4px padding from the previous 6px padding.

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -455,7 +455,10 @@ body {
     }
 
     &.neon-toolbar-thin {
-        padding: 4px !important;
+        padding: 6px !important;
+        height: 32px !important;
+        min-height: 32px !important;
+        line-height: 20px !important;
     }
 
     &.neon-toolbar-wide {
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -455,7 +455,7 @@ body {
     }
 
     &.neon-toolbar-thin {
-        padding: 6px !important;
+        padding: 4px !important;
     }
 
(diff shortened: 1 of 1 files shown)


ThomasSchellenbergNextCentury:
Hey @NBrockman 

So, I was reviewing this ticket, and I think that the task was a bit misleading.  (Sorry that I didn't catch this earlier.)
- I beleive they want the padding to be 6px.  At the time, I think the padding was 10px (if you notice, that's the padding of a normal mat-toolbar set in styles.scss), but, sometime between when this ticket was made and now, the toolbars were changed to use neon-toolbar-thin.  (Also, 6px is about 1/3rd smaller than 10px.)  Looking at the screenshot titled "Widget Header" emphasizes to me that they want the padding to be 6px.
- Also, looking at the screenshot titled "Widget Header", they want the overall height of the toolbar to be smaller, too.  Since the font size is 14px (which, from experience, I know to be about 20px high), plus the 6px padding on the top and bottom, the new height should be 20+6+6=32px (the current height is 40px).  You can fix this by adding height, min-height, and line-height properties to the neon-toolbar-thin styles (the line-height can be 20px, while the others can be 32px).

Let me know if you have any questions.  Sorry again for the confusion.  (Unfortunately this happens sometimes.)  Thanks!

NBrockman:
Thank you for the explanation. I made the changes. I believe everything should be correct now.

ThomasSchellenbergNextCentury (approved):

PR #257, opened 2019-06-14
THOR-1044-Gear-Validate-Number

NBrockman:
Setting WidgetFreeTextOptions to have ability to validate string or number.


NBrockman:
Is this the functionality you were looking for?

PR #259, opened 2019-06-17
THOR-1262-Apply-Gear-Fix

NBrockman:
Fixed Apply button to not be highlighted when no changes were made. The apply button will hide if changes are reverted (not if first fieldsConfig).

### DIFF
--- a/src/app/components/gear/gear.component.ts
+++ b/src/app/components/gear/gear.component.ts
@@ -377,9 +377,6 @@ export class GearComponent implements OnInit, OnDestroy {
      * @arg {string} bindingKey
      */
     public updateOnChange(bindingKey: string) {
-        if (typeof (this.modifiedOptions[bindingKey]) === 'undefined') {
-            return;
-        }
         // If the original binding key has been changed and added before
         if (this.originalOptions.access(bindingKey) !== undefined) {
             if (this.originalOptions.access(bindingKey).optionType === OptionType.NON_PRIMITIVE) {
@@ -388,6 +385,14 @@ export class GearComponent implements OnInit, OnDestroy {
                     return;
                 }
             }
+        }
+        if (_.isEmpty(this.modifiedOptions[bindingKey]) && !_.isEmpty(this.originalOptions[bindingKey])) {
+            this.changeMade = true;
+            return;
+        }
+        if (typeof (this.modifiedOptions[bindingKey]) === 'undefined' && _.isPlainObject(this.originalOptions[bindingKey])) {
+            this.changeMade = false;
+            return;
         }
         this.changeMade = true;
         // TODO THOR-1061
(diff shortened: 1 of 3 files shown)


src/app/components/gear/gear.component.ts
```diff
@@ -376,6 +377,18 @@ export class GearComponent implements OnInit, OnDestroy {
      * @arg {string} bindingKey
      */
     public updateOnChange(bindingKey: string) {
+        if (typeof (this.modifiedOptions[bindingKey]) === 'undefined') {
```
ThomasSchellenbergNextCentury: @NBrockman This causes a bug when you open the menu, set a value on a non-primitive field, click Apply, reopen the menu, and delete the value (the Apply button is still disabled, but it shouldn't be).

ThomasSchellenbergNextCentury:
@NBrockman FYI I found an easy way of testing this:  go into the Base Neon Component and, on this line, change "false" to "true":  https://github.com/NextCenturyCorporation/neon-dash-internal/blob/lorelei-master/src/app/components/base-neon-component/base-neon.component.ts#L1095

That should show a "Custom Widget Filter" option in each widget's gear menu.  Create a custom filter by setting the "lhs" (left-hand-side), "operator", and "rhs" (right-hand-side) properties, then click Apply and you should see that specific widget is filtered.

Here's an example:
```
lhs: location
operator: =
rhs: Turkey, Turkey
```

src/app/components/gear/gear.component.ts
```diff
@@ -376,6 +377,18 @@ export class GearComponent implements OnInit, OnDestroy {
...
+            return;
+        }
+        // If the original binding key has been changed and added before
+        if (this.originalOptions.access(bindingKey) !== undefined) {
+            if (this.originalOptions.access(bindingKey).optionType === OptionType.NON_PRIMITIVE) {
+                if (_.isEqual(this.originalOptions[bindingKey], this.modifiedOptions.access(bindingKey).valueCurrent)) {
```
ThomasSchellenbergNextCentury: @NBrockman FYI Rather than `this.modifiedOptions.access(bindingKey).valueCurrent` you can just say `this.modifiedOptions[bindingKey]`

NBrockman: @ThomasSchellenbergNextCentury Would you like me to make that change before you merge for improved readability?

ThomasSchellenbergNextCentury: Sounds good.

ThomasSchellenbergNextCentury:
@NBrockman This looks good, but you should add some unit tests to test the updateOnChange function with the three new cases you've added (the three "if" statements):  set a binding in the originalOptions and modifiedOptions, call updateOnChange with that binding key, and check that changeMade is the expected value.

ThomasSchellenbergNextCentury (approved):

PR #260, opened 2019-06-17
THOR Dashboard State Flow

arciisine:
Rework the state flow to allow for single path for all state changes, which will allow for url synchronization


src/app/dashboard/dashboard.component.ts
```diff
@@ -183,84 +166,97 @@ export class DashboardComponent implements AfterViewInit, OnInit, OnDestroy {
...
+            favicon.setAttribute('rel', rel);
+            favicon.setAttribute('type', 'image/x-icon');
+            favicon.setAttribute('href', icon);
+            head.appendChild(favicon);
+        }
+    }
 
-        title.innerText = this.projectTitle;
+    @DashboardModified()
+    private generalChange() {
```
ThomasSchellenbergNextCentury, line 250: This function is just here to trigger DashboardModified, right?

src/app/models/neon-namespaces.ts
```diff
@@ -15,10 +15,7 @@
 
 export namespace neonEvents {
     export const DASHBOARD_ERROR = 'DASHBOARD_ERROR';
-    export const DASHBOARD_READY = 'DASHBOARD_READY';
     export const DASHBOARD_REFRESH = 'DASHBOARD_REFRESH';
-    export const DASHBOARD_RESET = 'DASHBOARD_RESET';
-    export const DASHBOARD_STATE = 'DASHBOARD_STATE';
```
ThomasSchellenbergNextCentury, line 21: Please remove these events from the dashboard event guide:  https://github.com/NextCenturyCorporation/neon-dash-internal/blob/lorelei-master/DASHBOARD_EVENT_GUIDE.md

src/app/services/config.service.ts
```diff
@@ -18,21 +18,29 @@ import * as yaml from 'js-yaml';
...
 export class ConfigService {
     private configErrors = [];
-    private source = new ReplaySubject<NeonConfig>(1);
+    private source = new Subject<NeonConfig>();
 
     $source: Observable<NeonConfig>;
 
-    static as(config: NeonConfig) {
-        return new ConfigService(null, null).setActive(config);
+    static as(config: NeonConfig | null) {
```
ThomasSchellenbergNextCentury: The "as" function is just used for testing, so it seems weird to me to include it in the service itself.  I'd hate for someone to incorrectly use it in non-test code.

src/app/services/dashboard.service.ts
```diff
@@ -13,63 +13,74 @@
...
+        private searchService: AbstractSearchService
+    ) {
+        this.configSource = this.configService
+            .getActive()
+            .pipe(
+                mergeMap((config) => this.onConfigChange(config)),
+                shareReplay(1)
+            );
+        this.stateSource = this.stateSubject
+            .pipe(shareReplay(1));
```
ThomasSchellenbergNextCentury, line 60: Forgive my ignorance, but what's the difference between stateSubject and stateSource here?  Why is it useful to have both?

ThomasSchellenbergNextCentury:
@arciisine I'm getting lint errors running `npm run-script lint`

ThomasSchellenbergNextCentury:
@arciisine For our meeting this Friday on Observables, it would be great if you could speak about the RxJS functions you're using here, particularly in the Config and Dashboard Services, and how they work.  While we could all read the documentation ourselves, I think that having a short overview together as a team would also be nice.

src/app/dashboard/dashboard.component.ts
arciisine: Yes, its a catchall for any method that needs to modify the dashboard but doesn't have explicit functionailty

src/app/services/dashboard.service.ts
arciisine: Subject is writable, and is private as only the service should be able to emit events.  The source/observable is public and is read only.

src/app/services/config.service.ts
arciisine: Sounds like a good idea.

arciisine: Resolved, had to touch quite a bit, hopefully made it less brittle for the future.

arciisine:
Lint errors should be resolved, and all feedback should be addressed.

ThomasSchellenbergNextCentury (approved):

PR #261, opened 2019-06-18
THOR-1044-Gear-Validate-Nums

NBrockman:
Added a WidgetNumberOption and changed places in the code base that used those number options as hardcoded strings. This structure ensures number input from user on the gear option panel.


src/app/components/aggregation/aggregation.component.ts
```diff
@@ -487,10 +488,10 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
...
                 this.optionsTypeUsesGrid.bind(this)),
             new WidgetSelectOption('savePrevious', 'Save Previously Seen', false, OptionChoices.NoFalseYesTrue),
-            new WidgetFreeTextOption('scaleMinX', 'Scale Min X', '', this.optionsTypeUsesGrid.bind(this)),
-            new WidgetFreeTextOption('scaleMaxX', 'Scale Max X', '', this.optionsTypeUsesGrid.bind(this)),
-            new WidgetFreeTextOption('scaleMinY', 'Scale Min Y', '', this.optionsTypeUsesGrid.bind(this)),
-            new WidgetFreeTextOption('scaleMaxY', 'Scale Max Y', '', this.optionsTypeUsesGrid.bind(this)),
+            new WidgetNumberOption('scaleMinX', 'Scale Min X', 0, this.optionsTypeUsesGrid.bind(this)),
+            new WidgetNumberOption('scaleMaxX', 'Scale Max X', 100, this.optionsTypeUsesGrid.bind(this)),
+            new WidgetNumberOption('scaleMinY', 'Scale Min Y', 0, this.optionsTypeUsesGrid.bind(this)),
+            new WidgetNumberOption('scaleMaxY', 'Scale Max Y', 10, this.optionsTypeUsesGrid.bind(this)),
```
ThomasSchellenbergNextCentury: @NBrockman Did `null` not work here?  By default, we don't want to specify any scales, because we want to let the charting library determine them itself (which it should do already if they're not numbers).  Think about what happens, for example, if we have data that's between 0 and 1:  we're stretching the scale out to 10 (or 100, depending on the axis), leaving a lot of empty space.  Or think about what happens if we have negative numbers:  since the scales start at 0, you don't see the negative numbers at all!

src/app/models/widget-option.ts
```diff
@@ -154,6 +155,53 @@ export class WidgetFreeTextOption extends WidgetOption {
...
+            }
+            this._intermediateValue = this._intermediateValue || null;
+        }
+        return this._intermediateValue;
+    }
+
+    set intermediateValue(value: any) {
+        this._intermediateValue = value;
+        try {
+            this.valueCurrent = _.isInteger(_.toNumber(value)) ? this._intermediateValue : 0;
```
ThomasSchellenbergNextCentury: @NBrockman I don't think we want to use `isInteger` here because we want it to work with non-integer numbers, too, like floats/doubles.

src/app/components/aggregation/aggregation.component.ts
NBrockman: When putting null, the defaults are just 0 so everything just scales down and focuses on 0 since all the max and mins are set to 0. We can talk about this

ThomasSchellenbergNextCentury: @NBrockman Ah, OK; thanks for the explanation.  We can discuss it further this afternoon, but here's my initial impression:  I think the default value in the number validation code (in WidgetNumberOption) needs to be `null`, not `0`, because `0` has a specific meaning (like in this case, with the scales).  I think it's appropriate for widgets that have number options to check if they're null/undefined and use default values if that's the case.

ThomasSchellenbergNextCentury:
@NBrockman You should add some tests for the WidgetNumberOption verifying that the getter/setter for intermediateValue and getValueToSaveInBindings work as expected.  You can reference the existing tests for the WidgetNonPrimitiveOption as an example.

src/app/models/widget-option.ts
```diff
@@ -154,6 +155,53 @@ export class WidgetFreeTextOption extends WidgetOption {
...
+        super(OptionType.NUMBER, false, bindingKey, prettyName, valueDefault, undefined, enableInMenu);
+    }
+
+    get intermediateValue() {
+        if (this._intermediateValue === undefined) {
+            try {
+                const value = this.valueCurrent || this.valueDefault;
+                this._intermediateValue = _.isNumber(_.toNumber(value)) ? value : null;
+            } catch {
+                // Consume Error
```
rartiss55: Might be worth adding a console.error line to the catch blocks here.

NBrockman: Should I add console errors to the Non-Primitive options as well? They also currently just ignore the getter and setter errors.

ThomasSchellenbergNextCentury (approved):

PR #262, opened 2019-06-18
THOR-1023 URL State synchronization

arciisine:
Allows for dashboard state to be represented in the URL, and for the system to respond to those changes and affect the state accordingly.

Sample URL:
```
http://localhost:4200/?filter=⟦⟦’.ldc_uyg_jul_18.ui_out.topic’,’≈’,’Food’,’or’⟧,⟦’.ldc_uyg_jul_18.ui_out.topic’,’≈’,’Evacuation’,’or’⟧,⟦’.ldc_uyg_jul_18.ui_out.topic’,’≈’,’Medicine’,’or’⟧⟧
```


src/app/services/config.service.spec.ts
```diff
@@ -72,3 +79,112 @@ describe('Service: ConfigService', () => {
...
+
+                            }
+                        }
+                    }
+                }
+            }
+        }));
+    }
+
+    it('setActiveByURL loads the appropriate config (with filters) given a url and base path', (done) => {
```
ThomasSchellenbergNextCentury, line 115: @arciisine Would it be good to add a test on setActiveByURL with a base path but no filter?

src/app/util/config.util.ts
```diff
@@ -0,0 +1,152 @@
...
+        );
+        return regex;
+    }
+
+    static translate(data: string, keyMap: Record<string, string>): string {
+        const regex = this.buildMatcher(keyMap);
+        return data.replace(regex, (key) => keyMap[key]);
+    }
+
+    static visitDashboards<T>(dashboard: NeonDashboardConfig, handler: {
```
ThomasSchellenbergNextCentury, line 77: @arciisine It might be nice to add some tests for visitDashboards and filterDashboards.

arciisine: Visit dashboard is fairly well tested via the other tests, transitively.  Not sure what your take is on that.

ThomasSchellenbergNextCentury: Ah, OK.  Sounds good.

ThomasSchellenbergNextCentury:
@arciisine So it looks like, when you load a saved state with filters, the filters aren't actually applied.

arciisine:
Definitely an oversight, will be a simple fix to resolve.

ThomasSchellenbergNextCentury:
@arciisine If I load Neon with its default dashboard (`localhost:4200`), then change the dashboard (adding a `path` to the URL), then hit the "back" button, the URL changes (removing the `path`), but the dashboard doesn't change.  I think it's a specific issue with the default dashboard.

Could we auto-populate the URL with the default dashboard title?  So it would always have a `path`?

ThomasSchellenbergNextCentury:
@arciisine The other day we had discussed using a slash notation for the dashboard titles.  For example, rather than `http://localhost:4200/?path=situation_discovery.darpa60.nepal_elisa_discovery`, it would be `http://localhost:4200/situation_discovery.darpa60.nepal_elisa_discovery/`.  Did you have another idea in mind?  Or is that just out-of-scope of this ticket?

ThomasSchellenbergNextCentury:
@arciisine If I set a filter, then hit the "back" button, the filter is removed from the URL but not from the dashboard itself.

Edit:  This only happens when going "back" would remove the final filter.

ThomasSchellenbergNextCentury:
@arciisine When we load a saved state with a filter, is there a way we can auto-generate and append the correct `?filter=` to the URL?

ThomasSchellenbergNextCentury:
@arciisine FYI I like the current "back" button behavior (minus the bugs) allowing the user to iterate over each individual change to the filters.  I think that allowing the user to easily access recently-viewed dashboards can be handled with the creation of a "recent dashboards" menu, an upcoming ticket:  https://nextcentury.atlassian.net/browse/THOR-1014

src/app/util/config.util.ts
```diff
@@ -0,0 +1,152 @@
...
+ * limitations under the License.
+ */
+import { NeonDashboardConfig, NeonDashboardLeafConfig, NeonDashboardChoiceConfig } from '../models/types';
+
+export class ConfigUtil {
+    static encodeFiltersMap = {
+        '"': '’',
+        '=': '≈',
+        '[': '⟦',
+        ']': '⟧',
```
ThomasSchellenbergNextCentury, line 22: @arciisine Is there a way we could not use unicode characters in the URL encodings?  Ideally, I'd like to be able to write them on my own, without having to memorize unicode keyboard combinations.

arciisine: Per our discussion, plain JSON is still supported, just makes for messier URLs.

arciisine:
All feed back should be resolved

ThomasSchellenbergNextCentury (approved):

PR #268, opened 2019-06-20
THOR-1279 & THOR-1072-Separate-Layer-Options

NBrockman:
Separated Layer options into required and optional like other widgets gear panels look like. The the limit option is now always on the bottom near the buttons and has been removed from the 'Optional Config'  dropdown menu to prevent duplicate places of the limit field. The problem of the fields displaying in the correct order is also fixed. 
I would like to suggest a change in the naming of the layer options titles, as when the layer options appear there are many places in the options panel with 'Required Config' and 'Optional Config', making it hard to tell which is the layer config or the widget config.


rartiss55:
The THOR-1279 part looks fine.

For THOR-1072, it looks like the map layer options that used to exist after the Title, Database, and Table are missing now? I don't see where I would specify Latitude/Longitude fields, Color Field, Date Field, etc.

src/app/components/gear/gear.component.ts
```diff
@@ -299,6 +237,16 @@ export class GearComponent implements OnInit, OnDestroy {
...
+        // If (this.comp) {
+        //     this.originalOptions = this.comp.options;
+        //     /* eslint-disable-next-line @typescript-eslint/unbound-method */
+        //     this.exportCallbacks = [this.comp.exportData];
+        //     this.resetOptions();
+        //     this.constructOptions();
+        // }
+    }
+
+    ngAfterViewInit() {
```
NBrockman, line 217: I do not remember adding this at all. I do believe that this code is from an older pull from master, that has now been taken out. I do not believe that this should be added back, if it has previously already been removed. Would you like me to remove this? It does change functionality, as it changes the function names.

rartiss55: Yeah I think this block can be removed/was probably a mistake -- the ngOnInit logic that I see on lorelei-master starts on line 250, and I don't see an ngAfterViewInit() on lorelei-master.

ThomasSchellenbergNextCentury: I concur; you can remove ngOnInit.

src/app/components/options-section/options-section.component.ts
```diff
@@ -0,0 +1,130 @@
...
+    styleUrls: ['./options-section.component.scss'],
+    encapsulation: ViewEncapsulation.Emulated,
+    changeDetection: ChangeDetectionStrategy.Default
+})
+export class OptionsSectionComponent {
+    @Input() modifiedOptions: WidgetOptionCollection;
+    @Input() updateOnChange: Function;
+    @Input() handleChangeDatabase: Function;
+    @Input() handleChangeTable: Function;
+    @Input() index: number;
```
ThomasSchellenbergNextCentury: I think you can remove index; it doesn't look like you're still using it in this iteration.

src/app/components/gear/gear.component.ts
```diff
@@ -46,28 +46,18 @@ import { DashboardState } from '../../models/dashboard-state';
 export class GearComponent implements OnInit, OnDestroy {
```
ThomasSchellenbergNextCentury: When you remove ngOnInit, you also need to remove the OnInit implementation, or else it won't compile.

src/app/components/options-section/options-section.module.ts
```diff
@@ -0,0 +1,38 @@
...
+
+@NgModule({
+    declarations: [OptionsSectionComponent],
+    exports: [OptionsSectionComponent],
+    imports: [
+        MatToolbarModule,
+        MatFormFieldModule,
+        MatInputModule,
+        MatIconModule,
+        MonacoEditorModule.forRoot({ baseUrl: 'assets/' }),
```
ThomasSchellenbergNextCentury: We hadn't talked about this before, but, in a module file, you should only need to import things that are directly referenced in the component.  That means you should be able to remove things like MatInputModule and MonacoEditorModule from here.  Take a look, remove unneeded imports, and make sure `npm start` and `npm test` still run without errors.

src/app/components/gear/gear.component.ts
NBrockman: Yes I was actually making this change as you were making this comment. It is already pushed.

src/app/components/options-section/options-section.component.html
```diff
@@ -0,0 +1,50 @@
+<mat-toolbar class="subheader" color="primary">
+    <span>Required Config</span>
+</mat-toolbar>
+
+<!-- Non MultiLayer Support-->
+<div class="option-well">
+    <mat-form-field *ngIf=!modifiedOptions.layers.length class="option-container">
```
ThomasSchellenbergNextCentury: FYI While we haven't been completely uniform about it before, it's usually a good practice to have quotation marks around all HTML attributes (for example, here, having `*ngIf="!modifiedOptions.layers.length"`).  You always need quotation marks if the value contains spaces or HTML-reserved-characters, and it's nice to be consistent (using quotation marks for all attributes).

src/app/components/gear/gear.component.ts
```diff
@@ -166,6 +102,7 @@ export class GearComponent implements OnInit, OnDestroy {
         let optionList: WidgetOption[] = layer.list();
         optionList = this.removeOptionsByEnableInMenu(optionList, false);
         optionList = this.removeOptionsByBindingKey(optionList, 'title');
+        optionList = this.removeOptionsByBindingKey(optionList, 'limit');
         optionList = this.removeOptionsByType(optionList, 'DATABASE');
         optionList = this.removeOptionsByType(optionList, 'TABLE');
```
ThomasSchellenbergNextCentury: I forget if you already did this, but, as a reminder, you should be able to delete getLayerList and the removeOptions functions from the GearComponent.

ThomasSchellenbergNextCentury (approved):

PR #275, opened 2019-06-23
THOR-1311 Removed dependencies on DashboardService and DashboardState from ColorThemeService

ThomasSchellenbergNextCentury:
Removed dependencies on DashboardService and DashboardState from ColorThemeService (previously WidgetService).  The color sets are now re-initialized by the Dashboard Component.

### DIFF
--- a/src/app/models/types.ts
+++ b/src/app/models/types.ts
@@ -28,7 +28,7 @@ export interface NeonSimpleSearchFilter {
 
 export interface NeonDashboardOptions {
     connectOnLoad?: boolean;
-    colorMaps?: Record<string, any>;
+    colorMaps?: Record<string, Record<string, Record<string, Record<string, string>>>>;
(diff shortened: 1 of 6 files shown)


src/app/models/types.ts
```diff
@@ -28,7 +28,7 @@ export interface NeonSimpleSearchFilter {
 
 export interface NeonDashboardOptions {
     connectOnLoad?: boolean;
-    colorMaps?: Record<string, any>;
+    colorMaps?: Record<string, Record<string, Record<string, Record<string, string>>>>;
```
arciisine: Consider making this into a type, ColorMap, and reusing where needed.  It will be very easy for the type to get out of sync when used other places.

ThomasSchellenbergNextCentury: Done.

arciisine (approved):

NBrockman (approved):

PR #276, opened 2019-06-24
THOR-1310 Replaced FILTERS_CHANGED and FILTERS_REMOVED events

ThomasSchellenbergNextCentury:
Goals:
- Allow the visualizations and the FilterService to exist in an application without the DashboardComponent.

Results:
- Replaced the FILTERS_CHANGED and FILTERS_REMOVED events with custom listeners and notifiers in the FilterService.
- Each BaseNeonComponent registers its handleFiltersChanged listener function with the FilterService.
- The DashboardComponent overrides the default notifier with its onFiltersChanged function.
- On each filter, the FilterService calls DashboardComponent.onFiltersChanged, which calls FilterService.notifyFilterChangeListeners, which calls BaseNeonComponent.handleFiltersChanged.

### DIFF
--- a/DASHBOARD_EVENT_GUIDE.md
+++ b/DASHBOARD_EVENT_GUIDE.md
@@ -6,7 +6,6 @@ Event Channel | Intent | Event Payload
 --------------|--------|--------------
 DASHBOARD_ERROR | Shows an error message. | { error: Error, message: string }
 DASHBOARD_REFRESH | Signals a resize on all of the widgets. | {}
-FILTERS_CHANGED | Signals that one or more filters have been added or removed and the visualizations must requery. | { caller: string, change: Map }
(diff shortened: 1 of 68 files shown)


arciisine (commented):
My biggest request would be that if we are moving away from the event messenger, to use an observable as the event source instead of replicating the event messenger architecture inside of the FilterService.

ThomasSchellenbergNextCentury:
@arciisine Using an observable would introduce a dependency on RxJS, right?  While we could do that, it does give me some concern.

ThomasSchellenbergNextCentury:
@arciisine Using an observable would introduce a dependency on RxJS, right?  While we could do that, it does give me some concern.

arciisine:
For a service in the neon-framework I don't think that will be an issue, and for webcomponents, services don't even really exist (outside of a single component).

PR #280, opened 2019-06-25
THOR-1103 Map Bounds Redraw

arciisine:
Redraw map bounds when filter set programmatically

### DIFF
--- a/src/app/components/map/map.component.spec.ts
+++ b/src/app/components/map/map.component.spec.ts
@@ -23,10 +23,10 @@ import {
 
 import { MapComponent } from './map.component';
 
-import { AbstractSearchService } from '../../services/abstract.search.service';
+import { AbstractSearchService, CompoundFilterType } from '../../services/abstract.search.service';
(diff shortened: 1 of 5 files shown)


ThomasSchellenbergNextCentury:
@arciisine I don't think you'll be able to properly resolve this issue until THOR-1166-Dark-Theme is merged into lorelei-master since that affects how both the theme-change and the dashboard-refresh behavior is handled.

arciisine:
I don't think this relies on any of those changes, it only added on to an existing method to draw the boundary if the geo bounds are provided.  If that logic has changed, then I would concur this should be affected.

Konnoke:
I merged the Dark theme branch in order to test it out and it works!


ezgif com-video-to-gif

Konnoke (approved):

ThomasSchellenbergNextCentury:
So, I totally confused this task with THOR-1277 (the map zoom refresh bug), and thus my earlier comment.  Sorry about that.

Thanks to Kevin for the video!  Looks like everything is working properly.

PR #283, opened 2019-06-26
THOR Dashboard Choice as Path

arciisine:
The dashboard selection should now show up in the path vs a query parameter.  Also addresses some minor issues.

### DIFF
--- a/src/app/services/config.service.ts
+++ b/src/app/services/config.service.ts
@@ -147,7 +147,7 @@ export class ConfigService {
     /**
      * Loads config state by URL, given a base path
      */
-    setActiveByURL(url: string, base: string) {
+    setActiveByURL(url: string | Location, base: string) {
         const { filename, filters, dashboardPath } = ConfigUtil.getUrlState(url, base);
 
         setTimeout(() => {
--- a/src/app/services/config.service.ts
+++ b/src/app/services/config.service.ts
@@ -146,7 +146,7 @@ export class ConfigService {
      * Loads config state by URL, given a base path
      */
     setActiveByURL(url: string, base: string) {
-        const { filename, filters, path } = ConfigUtil.getUrlConfig(url, base);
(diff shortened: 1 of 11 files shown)


src/app/util/config.util.spec.ts
```diff
@@ -165,4 +165,27 @@ describe('Config Util Tests', () => {
...
     });
+
+    it('should properly parse URL states', () => {
+        // Ensure baseHREF works
+        for (const ctx of ['ctx', '/ctx', 'ctx/', '/ctx/']) {
+            const res = ConfigUtil.getUrlState('http://localhost/ctx/file/path1/path2?random#filters', ctx);
+            expect(res.dashboardPath).toEqual('path1.path2');
+            expect(res.pathParts).toEqual(['file', 'path1', 'path2']);
+            expect(res.filters).toEqual('filters');
+        }
```
ThomasSchellenbergNextCentury, line 176: Generally, I'm not a big fan of loops in tests, especially since it can be hard to identify failing assertions by line number.

src/app/services/config.service.ts
```diff
@@ -142,22 +147,18 @@ export class ConfigService {
...
-        const [path] = fullPath.split('/').filter((part) => !!part);
-        const params = decodeURIComponent(urlObj.hash.replace(/^#/g, ''));
-        const dashboardPath = urlObj.searchParams.get('path');
-        const cleanPath = path.replace(/^[/]+/g, '');
+    setActiveByURL(url: string | Location, base: string) {
+        const { filename, filters, dashboardPath } = ConfigUtil.getUrlState(url, base);
 
         setTimeout(() => {
-            from(cleanPath ? this.load(cleanPath) : throwError(null)).pipe(
+            // TODO: handle when we get rid of default
```
ThomasSchellenbergNextCentury: We might want to tag this TODO (and the one in ConfigUtil) with the corresponding JIRA ticket:  THOR-1300

PR #289, opened 2019-07-02
THOR-1278-Filter-Builder-Blank-Bug

NBrockman:
Fixed the filter builder from disappearing when valid queries would show no data.


src/app/components/aggregation/aggregation.component.ts
```diff
@@ -1423,7 +1424,7 @@ export class AggregationComponent extends BaseNeonComponent implements OnInit, O
      * @override
      */
     toggleBodyContainer() {
-        let bodyContainer = document.getElementsByClassName('body-container').item(0) as HTMLElement;
+        let bodyContainer = this.thisElementRef.nativeElement.getElementsByClassName('body-container').item(0) as HTMLElement;
```
ThomasSchellenbergNextCentury, line 1427: FYI You don't have to use thisElementRef; you can just use the existing "visualization" ElementRef.

PR #292, opened 2019-07-05
THOR-1011-About-Panel-Build-Date

NBrockman:
Added build date and latest commit in the About Panel. It will update the build date and last commit ID everytime 'npm run-script build' is run. Also fixed Neon logo to display from correct source.


rartiss55 (commented):
@NBrockman This is good! The only thing I'd say is that maybe we'd want to store this info in a local file that was ignored by git so that we wouldn't need to check in the environment file anytime someone does a build (also since LORELEI and AIDA do their builds differently/at different times, this info likely wouldn't match up between the two anyway). And if that local file wasn't created properly by the build script or doesn't exist, the UI could just display "N/A". @ThomasSchellenbergNextCentury what do you think?

package.json
```diff
@@ -4,7 +4,7 @@
   "license": "Apache2",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve --proxy-config proxy.conf.json  --host 0.0.0.0 --port 4200",
+    "start": "npm run build && ng serve --proxy-config proxy.conf.json  --host 0.0.0.0 --port 4200",
```
ThomasSchellenbergNextCentury: @NBrockman So I'm hesitant to run a complete build here since it will take a few minutes (more or less, depending on the build parameters) and seems like overkill.  I think it would be better for you to:

1. Move the environment-file-generation-code from build_dashboard.sh into a separate script file (maybe "generate_environment.sh").
2. Run generate_environment inside the build_dashboard script.
3. Run generate_environment inside "npm start" rather than calling "npm run build", so we don't run a complete build during development, like:  `"start": "./scripts/generate_environment.sh && ng serve ..."`

So "npm start" (developer build) runs generate_environment and "ng serve", and "npm build" (complete build) runs build_dashboard which runs generate_environment.

What do you think -- would that still work?

(Edit:  Updated wording)

NBrockman: Yep easy fix, makes sense.

ThomasSchellenbergNextCentury (approved):
Looks awesome!  I made a couple of minor changes, but overall you did a great job.

PR #294, opened 2019-07-08
THOR-1236 Filter Labels

ThomasSchellenbergNextCentury:
Goals:  Unify the code that generates the filter labels.

Accomplishments:
- Removed the code to generate filter labels from the Current Filters Component.
- Updated the code to generate filter labels in the Abstract Filter classes.  Compound Filters now generate specific labels if they correspond to specific designs like bounds, domain, or list.

### DIFF
--- a/src/app/components/aggregation/aggregation.component.spec.ts
+++ b/src/app/components/aggregation/aggregation.component.spec.ts
@@ -27,7 +27,7 @@ import { AbstractSearchService, CompoundFilterType } from '../../services/abstra
 import { InjectableColorThemeService } from '../../services/injectable.color-theme.service';
 import { AggregationType } from '../../models/widget-option';
 import { DashboardService } from '../../services/dashboard.service';
-import { CompoundFilterDesign, SimpleFilterDesign } from '../../util/filter.util';
+import { CompoundFilterDesign, FilterCollection, SimpleFilter, SimpleFilterDesign } from '../../util/filter.util';
 import { InjectableFilterService } from '../../services/injectable.filter.service';
 
 import { Color } from '../../models/color';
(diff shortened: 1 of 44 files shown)


arciisine:
The geo field name needs to be shortened, currently it shows as `(geoLocation.lat and geoLocation.lon)`, and should show as `geoLocation`.

ThomasSchellenbergNextCentury:
> The geo field name needs to be shortened, currently it shows as `(geoLocation.lat and geoLocation.lon)`, and should show as `geoLocation`.

@arciisine Respectfully disagree.
1. That won't work if the latitude and longitude fields aren't in the same parent object.
2. That's a bit misleading since the "geoLocation" object contains other properties.  We're not filtering on geoLocation.hash.

arciisine:
Also would recommend dropping the `=` from the label, as that should be indicated purely by it's presence.

arciisine:
Complex filters need to be truncated, `(entities.name and entities.translation and entities.translation)
(contains a and contains b and contains e)` is not clear and it takes up a lot of space.

arciisine:
> > The geo field name needs to be shortened, currently it shows as `(geoLocation.lat and geoLocation.lon)`, and should show as `geoLocation`.
> 
> @arciisine Respectfully disagree.
> 
> 1. That won't work if the latitude and longitude fields aren't in the same parent object.
> 2. That's a bit misleading since the "geoLocation" object contains other properties.  We're not filtering on geoLocation.hash.

I think it would be clear enough when looking at coordinate pairs.  Going for the 99% use case here, the extra space used in the filter bar is less than ideal.  If you are mixing/matching different fields for lat/lon then I would totally agree on showing both fields.

ThomasSchellenbergNextCentury:
@arciisine Thank you for the good suggestions.  I've:

1. Hidden the `=` operator.
2. Combined multiple filter fields that belong to common parent objects (like geoLocation).
3. Shortened labels of custom compound filters (though I'm sure we'll want to continue thinking about better ways to show them, as represented by ticket THOR-1333).

Please let me know if you have any more thoughts.

arciisine:
The custom filters still do not show the appropriate structure. `entities.end_char and entities.name
[ contains (a, ) and b ]` still looks confusing and wrong.

ThomasSchellenbergNextCentury:
@arciisine I've added a placeholder for empty string values.  For custom compound filters, I've moved the fields into the chip in order to make it look better.  Please let me know if you like it.

PR #296, opened 2019-07-08
THOR-1338 Fixed a bug with queries that aggregate by day in the Aggregation and Timeline components

ThomasSchellenbergNextCentury:
Previously, if you changed the date granularity to "day", you'd get no results.  That should work now.

### DIFF
--- a/src/app/components/aggregation/aggregation.component.spec.ts
+++ b/src/app/components/aggregation/aggregation.component.spec.ts
@@ -25,7 +25,7 @@ import { ChartJsScatterSubcomponent } from './subcomponent.chartjs.scatter';
 
 import { AbstractSearchService, CompoundFilterType } from '../../services/abstract.search.service';
 import { InjectableColorThemeService } from '../../services/injectable.color-theme.service';
-import { AggregationType } from '../../models/widget-option';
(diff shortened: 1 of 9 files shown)


gitnordt (commented):
I am getting either significant slowness or my browser locks up completely when I select January dates in the timeline for some reason. I am not getting a consistent error and I am not sure if you can recreate the issue on your end. I am just putting it on your radar in case you run into any issues or may know why it could be slow.

gitnordt (approved):

ThomasSchellenbergNextCentury:
@snorpet Thanks for putting it on my radar.  I can consistently reproduce the error by selecting a different date range, then re-brushing over January; it completely locks up my browser tab.  It happens with both "day" and "month" granularity, so I don't think it's directly related to this ticket.  I'm guessing that it's being caused by the act of replacing the existing set of filters with a new set, and because January has over 700 ID filters.  For now, you can avoid the issue by deleting the previous timeline filter using the "Remove All Filters" button in the navbar, then brushing over the January bars.  I've added a new ticket to further investigate the issue:  https://nextcentury.atlassian.net/browse/THOR-1342

PR #308, opened 2019-07-26
LORELEI Config File Update

ThomasSchellenbergNextCentury:
Added important fields to LORELEI config files.  (Due to a recent change, if any fields are defined in the config file, any fields that are NOT defined will be hidden from the dropdowns in the gear menus.) 
 Removed deprecated mapping options.  Fixed incorrect dataset relations.

### DIFF
--- a/src/app/config/config.TransitionPartner.yaml
+++ b/src/app/config/config.TransitionPartner.yaml
@@ -142,7 +142,7 @@ datastores:
                               prettyName: Date Created
                             - columnName: location
                               prettyName: Location
-                            - columnName: status
+                            - columnName: Status
                               prettyName: Status
                             - columnName: JustificationEN
                               prettyName: Original Text
@@ -189,6 +189,8 @@ datastores:
                               prettyName: Cluster Name
                             - columnName: cluster
                               prettyName: Cluster ID
+                            - columnName: type
+                              prettyName: Type
                             - columnName: location
                               prettyName: Location
                             - columnName: geoLocation.lat
@@ -362,6 +364,8 @@ datastores:
                     ui_output:
                         prettyName: UI out
                         fields:
+                            - columnName: createdAt
+                              prettyName: Date Created
                             - columnName: type
                               prettyName: Type
                             - columnName: location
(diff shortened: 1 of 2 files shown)


src/app/config/config.TransitionPartner.yaml
```diff
@@ -459,7 +571,7 @@ dashboards:
                                 text_cloud_data_key: es1.ldc_il3_meta.ui_out.location
                             relations:
                                 - - es1.ldc_il3_text.ui_output.topics
-                                  - es1.ldc_il3_meta.ui_output.topics
+                                  - es1.ldc_il3_meta.ui_output.topic
```
ThomasSchellenbergNextCentury, line 578: This looked like a bug since we define "topic" (singular) as group_field_key2.

src/app/config/config.TransitionPartner.yaml
```diff
@@ -499,12 +611,12 @@ dashboards:
...
-                                - - es1.ldc_il3_text.ui_output._id
-                                  - es1.ldc_il3_meta.ui_output._id
-                                - - es1.ldc_il3_text.ui_output.id
-                                  - es1.ldc_il3_meta.ui_output.id
+                                - - es1.elisa_il3_text.ui_output.topics
+                                  - es1.elisa_il3_meta.ui_output.topic
+                                - - es1.elisa_il3_text.ui_output._id
+                                  - es1.elisa_il3_meta.ui_output._id
+                                - - es1.elisa_il3_text.ui_output.id
+                                  - es1.elisa_il3_meta.ui_output.id
```
ThomasSchellenbergNextCentury, line 623: This looked like a copy-paste error.

Konnoke:
For TransitionPartner config, 
(1) the News Feed in the Somali Dashboard is blank. I don't think these config changes are the cause since II checked out master and it also seems to be blank on master. The New Feed for the Somali Dashboard on the ec2 is displaying things fine so sometime between when we last build the ec2 and now something changed that is causing just the News Feed to be blank in the Somali Dashboard.
(2) The Needs bar chart in the VODNO chart is blank
(3) The "Types", "Locations", and "Status" bar charts in the Central African Republic Dashboard are blank

For the Apan config, the geolocation seems incorrect for the Indonesia Flooding 2019 dashboard and ELISA APAN dashboard. As in no points are shown in the map widget for those two dashboard:

image

ThomasSchellenbergNextCentury:
@Konnoke Great job.  Thanks for checking everything.  I was able to reproduce all of the issues (except Types and Locations in the CAR dashboard; they were working, but Status wasn't) and have fixed them all.

PR #311, opened 2019-07-29
Aida 881 - Converting relation nodes into links

gitnordt:
Converted relation nodes into links, enhanced graph selection features, fixed media viewer height issue.


ThomasSchellenbergNextCentury:
@snorpet I'm seeing some failures in the network graph's unit tests.

PR #315, opened 2019-08-07
Update to Dockerfile.

rartiss55:
Added another proxy pass to nginx for the pipeline service. Not a high priority to merge, just want to make sure its in the repo in case someone besides me needs to rebuild the UI next week during the demo.


ThomasSchellenbergNextCentury (approved):
We can keep this for now, but I'm going to make a ticket to dynamically update the Dockerfile so we don't have to add locations for the specific program services.

PR #320, opened 2019-08-12
Thor 1352 Text-To-Url

NBrockman:
Urls are now clickable in the newsfeed and annotation viewer, the places where original text/translated text appears, which had plain text urls.

### DIFF
--- a/src/app/components/annotation-viewer/annotation-viewer.component.ts
+++ b/src/app/components/annotation-viewer/annotation-viewer.component.ts
@@ -107,6 +107,7 @@ export class AnnotationViewerComponent extends BaseNeonComponent implements OnIn
     public colorKeys: string[] = [];
     public indexInclusive: boolean;
     public offset = 0;
+    public url: string = '';
 
     constructor(
         protected colorThemeService: InjectableColorThemeService,
@@ -714,5 +715,22 @@ export class AnnotationViewerComponent extends BaseNeonComponent implements OnIn
 
     showLegendContainer(): boolean {
         return (!this.options.singleColor && this.colorKeys.length > 0);
+    }
+
+    removeUrl(text: string) {
+        return neonUtilities.removeUrl(text);
+    }
+
+    hasUrl(text: string) {
+        let urlPattern = /(?:(?:https?|ftp):\/\/)?(?:\S+(?::\S*)?@)?(?:(?!10(?:\.\d{1,3}){3})(?!127(?:\.\d{1,3}){3})(?!169\.254(?:\.\d{1,3}){2})(?!192\.168(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\x{00a1}\-\x{ffff}0-9]+-?)*[a-z\x{00a1}\-\x{ffff}0-9]+)(?:\.(?:[a-z\x{00a1}\-\x{ffff}0-9]+-?)*[a-z\x{00a1}\-\x{ffff}0-9]+)*(?:\.(?:[a-z\x{00a1}\-\x{ffff}]{2,})))(?::\d{2,5})?(?:\/[^\s]*)?/ig;
+        let prefixPattern = new RegExp('^(http|https)://');
+        let test = urlPattern.test(text);
+        if (test) {
+            this.url = neonUtilities.checkStringForUrl(text);
+            if (!prefixPattern.test(this.url)) {
+                this.url = 'http://' + this.url;
+            }
+        }
+        return test;
     }
 }
--- a/src/app/components/annotation-viewer/annotation-viewer.component.ts
+++ b/src/app/components/annotation-viewer/annotation-viewer.component.ts
@@ -722,6 +722,7 @@ export class AnnotationViewerComponent extends BaseNeonComponent implements OnIn
     }
 
     hasUrl(text: string) {
+        // eslint-disable-next-line max-len
         let urlPattern = /(?:(?:https?|ftp):\/\/)?(?:\S+(?::\S*)?@)?(?:(?!10(?:\.\d{1,3}){3})(?!127(?:\.\d{1,3}){3})(?!169\.254(?:\.\d{1,3}){2})(?!192\.168(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\x{00a1}\-\x{ffff}0-9]+-?)*[a-z\x{00a1}\-\x{ffff}0-9]+)(?:\.(?:[a-z\x{00a1}\-\x{ffff}0-9]+-?)*[a-z\x{00a1}\-\x{ffff}0-9]+)*(?:\.(?:[a-z\x{00a1}\-\x{ffff}]{2,})))(?::\d{2,5})?(?:\/[^\s]*)?/ig;
         let prefixPattern = new RegExp('^(http|https)://');
         let test = urlPattern.test(text);
--- a/src/app/components/annotation-viewer/annotation-viewer.component.ts
+++ b/src/app/components/annotation-viewer/annotation-viewer.component.ts
@@ -718,18 +718,10 @@ export class AnnotationViewerComponent extends BaseNeonComponent implements OnIn
         return (!this.options.singleColor && this.colorKeys.length > 0);
     }
 
-    removeUrl(text: string) {
-        return neonUtilities.removeUrl(text);
-    }
-
     hasUrl(text: string) {
-        // eslint-disable-next-line max-len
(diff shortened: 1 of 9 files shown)


src/app/components/news-feed/news-feed.component.ts
```diff
@@ -357,4 +360,26 @@ export class NewsFeedComponent extends BaseNeonComponent implements OnInit, OnDe
...
+
+    removeUrl(text: string) {
+        return neonUtilities.removeUrl(text);
+    }
+
+    hasUrl(text: string) {
+        // eslint-disable-next-line max-len
+        let urlPattern = /(?:(?:https?|ftp):\/\/)?(?:\S+(?::\S*)?@)?(?:(?!10(?:\.\d{1,3}){3})(?!127(?:\.\d{1,3}){3})(?!169\.254(?:\.\d{1,3}){2})(?!192\.168(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\x{00a1}\-\x{ffff}0-9]+-?)*[a-z\x{00a1}\-\x{ffff}0-9]+)(?:\.(?:[a-z\x{00a1}\-\x{ffff}0-9]+-?)*[a-z\x{00a1}\-\x{ffff}0-9]+)*(?:\.(?:[a-z\x{00a1}\-\x{ffff}]{2,})))(?::\d{2,5})?(?:\/[^\s]*)?/ig;
+        let prefixPattern = new RegExp('^(http|https)://');
+        let test = urlPattern.test(text);
```
ThomasSchellenbergNextCentury: @NBrockman Can't you just test for the URL pattern in `checkStringForUrl`?  If it doesn't exist, the function returns `null`, and you can ignore the update to `this.url`.  This way you don't have to repeat the `urlPattern` both here and in `checkStringForUrl`, AND you don't have to repeat the regex test code here and in the annotation viewer.

src/app/models/neon-namespaces.ts
```diff
@@ -75,6 +74,22 @@ export namespace neonUtilities {
         return itemToReturn;
     }
 
+    export function checkStringForUrl(text: string) {
+        // Need to use match operator and not RegExp.exec() because use of global flag
+        // eslint-disable-next-line max-len
+        let urlPattern = /(?:(?:https?|ftp):\/\/)?(?:\S+(?::\S*)?@)?(?:(?!10(?:\.\d{1,3}){3})(?!127(?:\.\d{1,3}){3})(?!169\.254(?:\.\d{1,3}){2})(?!192\.168(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\x{00a1}\-\x{ffff}0-9]+-?)*[a-z\x{00a1}\-\x{ffff}0-9]+)(?:\.(?:[a-z\x{00a1}\-\x{ffff}0-9]+-?)*[a-z\x{00a1}\-\x{ffff}0-9]+)*(?:\.(?:[a-z\x{00a1}\-\x{ffff}]{2,})))(?::\d{2,5})?(?:\/[^\s]*)?/ig;
```
ThomasSchellenbergNextCentury: @NBrockman Move the `urlPattern` to a `const` in `neonUtilities` so you only have to define it once.

src/app/models/neon-namespaces.spec.ts
```diff
@@ -370,4 +370,14 @@ describe('NeonUtilities', () => {
             { keyA: 'value4', keyB: 'bravo' }
         ]);
     });
+
+    it('checkStringForUrl returns the first occurrence of a url in the string', () => {
+        let testString = 'Hello World, https://www.google.com Goodbye world';
```
ThomasSchellenbergNextCentury: @NBrockman We should be testing other links that the URL pattern should match as well, including (but not limited to):
- Links that start with `http://` or `ftp://`
- Links that don't start with `https` or `http` or `ftp`
- Links with and without `www`
- Links that end with `.net` or other domains
- Links that end with a query string (https://en.wikipedia.org/wiki/Query_string)
- Links that end with a fragment (https://en.wikipedia.org/wiki/Fragment_identifier)

src/app/models/neon-namespaces.ts
```diff
@@ -75,6 +74,22 @@ export namespace neonUtilities {
...
+        // eslint-disable-next-line @typescript-eslint/prefer-regexp-exec
+        let matches = text.match(urlPattern);
+        return matches[0];
```
ThomasSchellenbergNextCentury: @NBrockman The code should work with text that contains multiple links.  Your code only works with text containing a single link.  Please modify as needed.

src/app/components/news-feed/news-feed.component.html
```diff
@@ -29,15 +29,33 @@
...
+                        </p>
                     </div>
                     <div class="text multiple-line" *ngIf="item[options.secondaryContentField.columnName]">
                         <h4 *ngIf="options.secondaryContentLabel">{{options.secondaryContentLabel}}</h4>
-                        <p>{{item[options.secondaryContentField.columnName]}}</p>
+                        <p>
+                            <ng-container *ngIf="hasUrl(item[options.secondaryContentField.columnName])">
+                                {{removeUrl(item[options.secondaryContentField.columnName])}}
+                                <a *ngIf="hasUrl(item[options.secondaryContentField.columnName])" href="{{this.url}}" target="_blank">
+                                    {{this.url}}</a>
```
ThomasSchellenbergNextCentury: @NBrockman Your code is moving the URL to the end of the text, but I don't think that's what we want.  For example, the text `Go to google.com to find us!` would be shown as `Go to to find us! <a>google.com</a>` which is not great.  It should look like `<span>Go to</span><a>google.com</a><span>to find us!</span>`.  The Annotation Viewer does something like this with its `parts` that each represent word phrases (or "parts" of sentences).  Please take a look at it and update your code as needed to fix the issue.

src/app/components/news-feed/news-feed.component.html
```diff
@@ -29,15 +29,33 @@
                     <span class="text">{{ item[options.titleContentField.columnName] || item[options.contentField.columnName] }}</span>
                 </mat-panel-title>
             </mat-expansion-panel-header>
-            <mat-panel-description>
+            <mat-panel-description *ngIf="isExpanded(item)">
```
ThomasSchellenbergNextCentury, line 32: @NBrockman Why is `isExpanded` needed here?  It seems to be making the collapse animations a bit choppier.

NBrockman: I added that so that the logic of all the RegEx would not have to run on all 100 (or whatever the limit is set to) as it greatly slows down the load time of the widget. Instead it only loads and does computation when the expanded part of the accordion is actually expanded. It was my way of lazy-loading when the thing I want lazy loaded wasn't exactly one of our own components.  I implemented this in the media viewer ticket as well for this same purpose but the effect was exponentially more impactful, because with all the media in some of the datasets the component would take so long it would actually time out.

ThomasSchellenbergNextCentury:
@NBrockman I found two bugs in the Uyghur July 2018 dataset, 5th item in a news feed sorted by date ascending.  The `secondaryContent` is supposed to say:

> Live: There was a magnitude 6.6 earthquake in Xinjiang Uyghur Autonomous Region and it affected Urumqi strongly, June 30 (Alkuyi.net): June 30, 2012, 3:07 PM, there was magnitude 6.6 M earthquake in Xinyuan Town and Hejing Town, Yili Autonomous Prefecture, and Xinjiang Uyghur autonomous region (location: 43.4 N; 84.8 E). The seismic depth was up to 7 km.

This is the result:

Thor-1352-Bug

I see two separate bugs here:
1. Part of the `secondaryContent` is being added to the main `content` (the translated text is being added to the untranslated text).
2. The `secondaryContent` is being cut-off after the link.

Here's a copy of the `newsFeed` config you can use:

```yaml
        - type: newsFeed
          col: 1
          row: 15
          sizex: 6
          sizey: 8
          bindings:
              title: Document Feed
              tableKey: table_key_1
              idField: id_field_key
              linkField:
              dateField: date_field_key
              titleContentField: group_field_key
              contentField: original_text_key
              secondaryContentField: translated_text_key
              filterField: id_field_key
              sortField: date_field_key
              ignoreSelf: true
              limit: 100
```

src/app/components/annotation-viewer/annotation-viewer.component.ts
```diff
@@ -715,4 +717,19 @@ export class AnnotationViewerComponent extends BaseNeonComponent implements OnIn
     showLegendContainer(): boolean {
         return (!this.options.singleColor && this.colorKeys.length > 0);
     }
+
+    removeUrl(text: string) {
```
ThomasSchellenbergNextCentury: @NBrockman The Annotation Viewer, News Feed, and Neon Utilities all define `removeUrl` functions.  However, the functions in the Annotation Viewer and the News Feed are never called, and the function in Neon Utilities is only called by the functions in the Annotation Viewer and the News Feed (which, again, are never called themselves).

Assuming that the `removeUrl` functions truly aren't needed (I'm sure they were, in a previous iteration, but not any more), please remove all three of them.

src/app/components/news-feed/news-feed.component.html
```diff
@@ -29,15 +29,39 @@
...
                 <div>
                     <div class="text multiple-line" *ngIf="item[options.contentField.columnName]">
                         <h4 *ngIf="options.contentLabel">{{options.contentLabel}}</h4>
-                        <p>{{item[options.contentField.columnName]}}</p>
+                        <p>
+                            <ng-container *ngIf="hasUrl(item[options.secondaryContentField.columnName])">
```
ThomasSchellenbergNextCentury: @NBrockman I think I found one of your issues:  the `secondaryContentField` here should be `contentField`.

src/app/components/annotation-viewer/annotation-viewer.component.ts
```diff
@@ -715,4 +717,19 @@ export class AnnotationViewerComponent extends BaseNeonComponent implements OnIn
...
+        return neonUtilities.removeUrl(text);
+    }
+
+    hasUrl(text: string) {
```
ThomasSchellenbergNextCentury, line 721: @NBrockman The `hasUrl` function in both Annotation Viewer and News Feed seem to be doing the same thing, right?  If so, could we just move it to neonUtilities?  Not only would it consolidate the two functions, but you also wouldn't need two sets of identical tests (one in each component).

src/app/components/news-feed/news-feed.component.html
ThomasSchellenbergNextCentury: That makes sense.  Thanks for the thorough explanation.  It seems reasonable to keep the `ngIf` on `isExpanded` here for now.

PR #324, opened 2019-08-21
Aida 997

gitnordt:
Modify thumbnail grid code to handle configuring urls and fixed error for canvas not existing


ThomasSchellenbergNextCentury:
@snorpet I'm seeing some unit test errors.

PR #328, opened 2019-08-23
THOR-1042-Dynamic-Pretty-Names

NBrockman:
Allowed for pretty names to be a callback function for a dynamic naming convention


ThomasSchellenbergNextCentury (commented):
@NBrockman So, in `options-list-component.html`, where it shows the `prettyName`, can please you update the code so that, if the `prettyName` is a `Function`, it calls the function?

ThomasSchellenbergNextCentury (approved):

PR #329, opened 2019-08-27
THOR-1398 Save filter IDs and relations in URL filters.

ThomasSchellenbergNextCentury:
Fixed the bug where, if you remove a filter that has a related filter (for example, by clicking on the X in the Query Bar), the related filter won't be removed too.  I've added the filter ID and the relations to the URL filters.  I've also shortened the length of the filter IDs and the URL filter fragments.


gitnordt (approved):
The query bar still doesn't clear when all filters are removed, but I believe that's a separate issue so I am approving this PR.

gitnordt:
> The query bar still doesn't clear when all filters are removed, but I believe that's a separate issue so I am approving this PR.

I may have a fix for this. I will submit a separate PR : AIDA-754

PR #338, opened 2019-09-09
Wiki viewer update

Konnoke:
Added support for both Wikipedia ID and page name API

For Demo. Failed unit tests commented out.


ThomasSchellenbergNextCentury:
@Konnoke I fixed the unit tests.  There were two main issues:
- Remember when we had removed the `body` property from the `wikiResponse` object?  Well, we forgot to remove it from the tests.  Oops.
- The `links` were incorrectly typed as `string[]`, but they are actually `string|string[]`.  Some tests were failing when you tried to call `forEach` on what was actually a `string`.

src/app/components/wiki-viewer/wiki-viewer.component.ts
```diff
@@ -233,8 +236,14 @@ export class WikiViewerComponent extends BaseNeonComponent implements OnInit, On
     ): void {
         new Promise<number>((resolve, reject) => {
             try {
-                let links: string[] = neonUtilities.deepFind(results[0], options.linkField.columnName) || [];
-                this.retrieveWikiPage((Array.isArray(links) ? links : [links]), [], (data: WikiData[]) => {
+                let links = neonUtilities.deepFind(results[0], options.linkField.columnName) || [];
+                links = (Array.isArray(links) ? links : [links]).map((link) => {
+                    if (link.includes('https://en.wikipedia.org/wiki/')) {
```
ThomasSchellenbergNextCentury: @Konnoke I'd recommend adding a unit test on `handleTransformationVisualizationQueryResults` verifying this new behavior of modifying wikipedia links.

src/app/components/wiki-viewer/wiki-viewer.component.ts
```diff
@@ -272,17 +281,15 @@ export class WikiViewerComponent extends BaseNeonComponent implements OnInit, On
             this.retrieveWikiPage(links.slice(1), data, callback);
         };
 
-        this.http.get(WikiViewerComponent.WIKI_LINK_PREFIX + links[0]).subscribe((wikiResponse: any) => {
+        let link = (this.options.useWikipediaPageID ? WikiViewerComponent.WIKI_LINK_PREFIX_ID :
```
ThomasSchellenbergNextCentury, line 285: @Konnoke You can also add a unit test on `handleTransformationVisualizationQueryResults` with `useWikipediaPageID` set to `true`.

src/app/components/wiki-viewer/wiki-viewer.component.ts
ThomasSchellenbergNextCentury: @Konnoke FYI if you make `'https://en.wikipedia.org/wiki/'` a static variable then you can use `variableName.length` rather than relying on the magic number `30`.

ThomasSchellenbergNextCentury (approved):

PR #339, opened 2019-09-11
Adds prefix to /wiki/

Konnoke:
For Demo


ThomasSchellenbergNextCentury (approved):

rartiss55:
Meant to add this comment earlier -- I already let Chris/Todd/Sophie know for the demo tomorrow, but there a couple edge cases here that don't match the find + replace done on <a href="/wiki/...>
- Edit links
- Anchor links in the table of contents for a wikipedia page

I don't think this matters too much for the demo since I don't see anyone clicking "edit" and the anchor links at least won't give a 404 error if clicked. However, wanted to bring it up since I think this would be a different ticket and require discussion on how we want to approach these cases/what the wikiviewer should actually do (perhaps we want to ignore or disable the edit links for instance?)

ThomasSchellenbergNextCentury:
@rartiss55 Cool, I created THOR-1443 to tackle that.

PR #341, opened 2019-09-12
AIDA-754-fixed issue with the query bar not clearing when filters are removed

gitnordt:


src/app/components/query-bar/query-bar.component.ts
```diff
@@ -287,6 +288,12 @@ export class QueryBarComponent extends BaseNeonComponent {
      */
     refreshVisualization() {
         this.changeDetection.detectChanges();
+
+        let filters = this.filterService.getFilters();
```
ThomasSchellenbergNextCentury, line 292: FYI You can also put this in QueryBarComponent.redrawFilters and call getFilters() on the FilterCollection argument to verify if the QueryBar is filtered.  The redrawFilters function is called whenever the filters are changed.

ThomasSchellenbergNextCentury:
@snorpet This is good, but it only tackles half of the intent of the ticket:  removing a filter from outside the query bar now removes the text from the query bar (yay!), but adding a filter on the query field from outside the query bar (like, say, the filter builder) does not update the text in the query bar.  If you'd like to split this task into a separate ticket, please update the TODO in QueryBarComponent.redrawFilters to use the new ticket number.

gitnordt:
> @snorpet This is good, but it only tackles half of the intent of the ticket: removing a filter from outside the query bar now removes the text from the query bar (yay!), but adding a filter on the query field from outside the query bar (like, say, the filter builder) does not update the text in the query bar. If you'd like to split this task into a separate ticket, please update the TODO in QueryBarComponent.redrawFilters to use the new ticket number.

MY apologies. I misunderstood the ticket description. I have put in a new ticket to capture the issue AIDA-1041

ThomasSchellenbergNextCentury (approved):
My apologies as well -- I'll try to add better descriptions to my tickets in the future.

PR #343, opened 2019-09-16
Lessen Complicated Behavior in the FilterService:  Remove toggleFilters, Other Cleanup

ThomasSchellenbergNextCentury:
- Separated complex behavior in `FilterService` functions into smaller private functions, including `_createFilterAndRelations`, `_deleteFilterIds`, `_findFilterIds*`, and `_saveOrRetrievePreviousFilters`.
- Removed `FilterService.toggleFilters`.  Many visualizations now save an array (or Map) of filtered values so they know whether to toggle selected values ON or OFF (if toggling isn't disabled).  Added `CoreUtil` functions `changeOrToggle*`, `isItemFilteredInEveryField`, `retrieveValuesFromListFilters`, and `updateValuesFromListFilters` to handle shared behavior.
- Added `FilterService.createFilters` specifically to be called by the `FilterBuilder` component.
- Merged `FilterService.getRawFilters` with `FilterService.getFilters`.
- Moved `createRelationFilterList` from `FilterService` into `AbstractFilter`.
- Removed `FilterService.isFiltered` because most visualizations can now handle this behavior by themselves.
- Changed visualizations that created both `ListFilterDesign` and `SimpleFilterDesign` to just create `ListFilterDesign` since they are mostly redundant with one another.
- Added more comments everywhere.


src/app/components/taxonomy-viewer/taxonomy-viewer.component.ts
```diff
@@ -563,41 +554,50 @@ export class TaxonomyViewerComponent extends BaseNeonComponent implements OnInit
...
+        const subTypeFilter: FilterConfig = (subTypeValues.length && !subTypeIsDuplicated) ? this.createFilterConfigOnList(
+            this.options.subTypeField, subTypeValues
+        ) : null;
 
         // If we don't need to filter a valid filterable field, ensure that we delete all previous filters that were set on that field.
         const filterConfigListToDelete: FilterConfig[] = [];
         if (!categoryFilter && this.options.categoryField.columnName) {
-            filterConfigListToDelete.push(this.createFilterConfig(this.options.categoryField));
+            filterConfigListToDelete.push(this.createFilterConfigOnList(this.options.categoryField));
         }
```
gitnordt, line 594: I am seeing weird behavior with the taxonomy filtering. Steps to reproduce error:
1. Uncheck one category(e.g. EVENTS)
2. Uncheck a second category (e.g. ENTITIES)
3. Check and uncheck multiple types and/or sub-types in one of the categories that were unchecked
Incorrect behavior: The category filter toggles on and off with every type check action.

src/app/components/network-graph/network-graph.component.ts
```diff
@@ -1362,24 +1374,35 @@ export class NetworkGraphComponent extends BaseNeonComponent implements OnInit,
...
             } else {
-                this.exchangeFilters(filters);
+                // If we won't add any filters, create a FilterDesign without a value to delete all the old filters on the filter field.
+                filtersToDelete.push(this.createFilterConfigOnList(nodeFilterField.field));
             }
-        }
+        });
+
+        this.exchangeFilters(filters, filtersToDelete);
     }
```
gitnordt, line 1406: This is really just an observation. 

I am still unsure that the network graph filtering capability is instinctive enough. Currently, when you double-click on one node and then another, the filters change (are exchanged) which I believe is the expected behavior. However, if you click on the same node twice or click on a node that has equivalent filters to the previous node clicked, the filter is removed. This seems like it may be confusing to a user that doesn't understand the intent. To a user, it could look like double-clicking on nodes add a filter at times and other times delete a filter without reason. I wonder if it would be more intuitive to only remove network graph filters from the filter menu and not by double-clicking on nodes.

ThomasSchellenbergNextCentury:
@snorpet 
1. I fixed the bug you found in the taxonomy viewer.
2. I changed the filter behavior in the network graph to what you described by using the `keepSameFilters` feature you added a few weeks ago.  If you ever need to change it back, you can just remove the last argument (`true`) from the `exchangeFilters` call.  :)

PR #349, opened 2019-10-08
bug fix: download attributes fix for firefox

aamduka:


src/app/components/export-control/export-control.component.ts
```diff
@@ -65,11 +65,14 @@ export class ExportControlComponent {
 
     exportSuccess(queryResults) {
         let link = document.createElement('a');
-        link.href = 'data:text/csv;charset=utf-8,' + encodeURI(queryResults.data);
+        var url = URL.createObjectURL(new Blob([queryResults.data], {type:"application/octet-stream"}));
```
chrislong: `application/octet-stream`? Isn't this text data?

chrislong (changes requested):
Just wondering about the MIME type.

src/app/components/export-control/export-control.component.ts
aamduka: Chris,

I was thinking some browsers might try to display the content instead of prompting user to download when they see "text/plain" MIME type. With application/octet-stream no browser would attempt to open it. But all the browsers on my machine didn't try to open it when using "text/plain" so I went ahead with "text/plain" hoping older browsers would not attempt to open the file.

chrislong (approved):

PR #354, opened 2019-10-18
Thor 1463 quick installer for neon

aamduka:
Chris,

A new subfolder (quick-install) is created under neon-dashboard repo that contains all the files needed and a documentation of the steps to follow to create the install tarball.


quick-install/readme.md
```diff
@@ -0,0 +1,3 @@
+- install docker for your platform: https://docs.docker.com/v17.09/engine/installation/
```
ThomasSchellenbergNextCentury, line 1: @aamduka Sorry to nitpick, but this file should probably be named `README.md` to match the existing naming convention of documentation files.

ThomasSchellenbergNextCentury:
@aamduka The `docker-compose.yml`, `load.sh`, and `nginx.conf` files in the `quick-install` folder are almost exactly the same as the ones in the `e2e` folder.  Would it make sense to combine them somehow?  I know the e2e containers are named using an `e2e-` prefix, but I don't think that's really needed, as long as we update the names everywhere they're referenced.  We'd need to change the load script to accept a variable input as the dataset name, but the two will probably eventually use the same dataset whenever I get around to finishing THOR-1367.

aamduka:
@ThomasSchellenbergNextCentury That makes sense. Will try to combine the two. is moving the 'docker' folder one level up and sharing the docker config that way makes sense?

ThomasSchellenbergNextCentury:
@aamduka That seems right.  We'll want to make sure that running `./e2e.sh` still works.  (Since Jenkins is currently having issues, we can worry about verifying it still works there later).

ThomasSchellenbergNextCentury:
@aamduka As we already discussed this morning, I went ahead and replaced the ldc_uyg_jul_18 dataset in the E2E tests with the earthquakes dataset.  This should make combining the E2E test setup with the quick-install much easier (the load scripts should be the same now).  Also, I made a minor update to the load script (adding a limit flag to the elasticdump command) because I noticed that the E2E tests would start running before the earthquake data was completely loaded otherwise.

aamduka:
@ThomasSchellenbergNextCentury looks great. I will go ahead and merge your updates with my local.

aamduka:
@ThomasSchellenbergNextCentury I have made all the updates to share the docker staff between e2e and quick install. I figured that you can actually parameterize settings in docker compose using environment variables and I am able to use that to parametrize dist folder location for e2e and quick install.

@chrislong I am able to incorporate all your suggestions with some change to your suggestion on the install script. It turns out that "docker-compose pull" fails if any of your images are not registered on docker-hub. instead of checking both the local and the remote repos for images, it only checks the remote one (docker-hub). this is weird because "docker-compose up" actually checks for both locations. since one of our images (neon-server image) is located only in the local, it is going to be problematic to use "docker-compose pull". for this reason, I am only loading neon-server tarball image to the local docker repo inside the install script. the start script takes care of pulling all the remote images and starting the container.

chrislong:
@aamduka I didn't realize that about `docker-compose pull`. Ok, sounds good.

chrislong:
Just one nitpick: shell scripts should start with `#!/bin/sh` (at least ones that might be directly executed rather than `source`d, which I think is all of these).

quick-install/start.sh
```diff
@@ -0,0 +1,9 @@
+#installs and starts the neon system
+
+./install.sh
```
chrislong: It's safer to call a sibling shell script like this:

```
mydir=$(readlink --canonicalize "$(dirname $0)")
"${mydir}"/install.sh
```

Then it doesn't matter what the current directory is when it's run.

aamduka:
@chrislong I could not get the startup.sh script to work using "readlink --canonicalize" on Mac. I used a different technique to basically do the same thing. it now works when launched from any directory.

chrislong (approved):

PR #357, opened 2019-10-23
Moved LORELEI Program and Performer Information to Dash Config File

ThomasSchellenbergNextCentury:
- Moved the LORELEI program and performer information to the dashboard config file
- Updated all recent LORELEI config files
- Added the Neon Server's build date and git commit ID to the About panel
- Wrote some unit tests for the About component
- Minor changes to the Neon Dashboard "about" text

Please run "npm install" before testing.


aamduka (commented):
@ThomasSchellenbergNextCentury I was wondering if we should move the html stuff (tags, styles, etc) in the config files to the about component and only keep the program and performer parameters in the config files. (eg. program-description, program-link, performers-list, etc). what do you think of that?

ThomasSchellenbergNextCentury:
@aamduka Good idea.  I've updated the code to accept either a string (like before) or an object with a specific set of configurable properties.  I've updated the LORELEI config files and added examples to the Dashboard Configuration Guide.  Let me know what you think!

aamduka:
@ThomasSchellenbergNextCentury Looks good.

aamduka (approved):

PR #359, opened 2019-11-01
Thor 1255 angular upgrade to 8.2.9

aamduka:
- Angular 8 upgrade required other packages to be upgraded as well 
- Typescript upgraded to 3.5.3. this is stricter than the previous version which required some code 
   to be manually updated
- wishtack reactive component loader upgraded to 1.1.0. because angular 8 changed the signature 
   of the interface used for lazy loading components
- core-js upgraded to 3.3.6 requiring some polyfills to be updated which in turn required some 
  manual changes to the code to fix a few failing unit tests


ThomasSchellenbergNextCentury:
@aamduka Everything seems to be working well, but I had a question about the changes to `ViewChild`:  Is there a reason we're setting `static` to `true` everywhere?  It looks like `false` is the default, and `true` is only intended for specific use cases:  https://stackoverflow.com/a/56359612

Is there a case in the Dashboard in which setting `static` to `false` causes an error?

ThomasSchellenbergNextCentury:
@aamduka I'm getting lint errors.  Take a look with `npm run-script lint`

I think most of the errors are due to the change to `::ng-deep` in the scss files.

aamduka:
> https://stackoverflow.com/a/56359612

@ThomasSchellenbergNextCentury there are actually a few cases where static is set to false in our code. examples are @ViewChild('table', { static: false }) and @ViewChild('dragView'). but in most of the cases it is set to true because most of the @ViewChilds in our code don't have binding resolution so it is safe to set static true. ng update tool is smart enough to detect that viewchild should be set static or not automatically for most cases.

aamduka:
> @aamduka I'm getting lint errors. Take a look with `npm run-script lint`
> 
> I think most of the errors are due to the change to `::ng-deep` in the scss files.

From angular upgrade guide: "/deep/ and ::ng-deep both are deprecated but using ::ng-deep is preferred". Will take a look at the lint issue

aamduka:
@ThomasSchellenbergNextCentury reverted back to /deep/ to fix the lint errors. according to angular docs (https://angular.io/guide/component-styles), both ::ng-deep and /deep/ are deprecated and would be removed soon (I suspect probably in angular 9). we may have to revisit this issue when that happens.

ThomasSchellenbergNextCentury:
@aamduka 

Re. `static` `false` vs `true`: Thank you for the explanation!

Re. deprecation of `/deep/` selectors: We'll watch out for it, but I'd imagine that they'll need to replace the existing functionality with an alternative.  Otherwise a lot of our custom styling will need to be fixed...

ThomasSchellenbergNextCentury (approved):

PR #360, opened 2019-11-04
THOR-1478 THOR-1479 Various Data Table Updates

ThomasSchellenbergNextCentury:
- Added a linkFields option to the data table to show links in table cells.
- Fixed changing the showFields in the option menu.
- Fixed updating column widths on resize.
- Updated the data table styles.
- Removed deprecated data table functions.
- Misc. cleanup.


aamduka (approved):

ThomasSchellenbergNextCentury:
@dgarcia-bot Good question.  We used to have an option to choose between two slightly different table styles, but that option's no longer needed, so I've merged them.

PR #365, opened 2019-11-13
Thor 1465: import CSV functionality

aamduka:
- added a new component for selecting a CSV file, parsing into JSON and import the data into a selected data table
- used third party JS library to parse CSV into JSON
- added a menu item called (import CSV) in the hamburger menu located on top right corner. menu item opens the CSV import screen on which users select source file and destination store and initiate the import


ThomasSchellenbergNextCentury:
FYI @chrislong and other reviewers:  Since the middleware changes are still on a feature branch, you have to update the `package.json` in the Neon Dashboard and add the branch name (THOR-1465) to the end of the neon-framework dependency:

```
"neon-framework": "git+https://github.com/NextCenturyCorporation/neon-framework.git#THOR-1465",
```

Then rerun `npm install`

src/app/components/import-data/import-data.module.ts
```diff
@@ -0,0 +1,40 @@
...
+        MatDividerModule,
+        MatRadioModule,
+        MatButtonModule,
+        MatSelectModule,
+        MatFormFieldModule,
+        FormsModule,
+        CommonModule
+    ],
+    providers: [
+        { provide: Object, useValue: papaparse }
```
ThomasSchellenbergNextCentury: @aamduka Out of curiosity, why do you import papaparse in the Module and then inject it into the Component, rather than just importing it directly in the Component file?  Is there a particular advantage to doing it this way (or is this a best practice), or is this just personal preference?

src/app/components/import-data/import-data.component.ts
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    public onCancel() {
+        this.sideNavRight.close();
+    }
+
+    public onImportClick() {
+        let file = this.inputFile.nativeElement.files[0];
+        this.papa.parse(file, { header: true,
+            complete: this.import.bind(this) });
```
ThomasSchellenbergNextCentury: @aamduka You should add `skipEmptyLines: true` to the parse options here.

src/app/components/import-data/import-data.component.html
```diff
@@ -0,0 +1,45 @@
+<form class="gear-container">
```
ThomasSchellenbergNextCentury, line 1: @aamduka We probably want to add a sentence about requiring the CSV to have a header row.

Akmal and @chrislong are there any other instructions for the user we should add here?

src/app/components/import-data/import-data.component.ts
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    import(result: any) {
+        if (result.errors.length > 0) {
+            this.csvParseError = `CSV parsing error at line ${result.errors[0].row}: ${result.errors[0].message}`;
+            return;
+        }
+        else
+        {
+            this.csvParseError = "";
```
ThomasSchellenbergNextCentury: @aamduka You don't need this here since you're already doing it on line 114.

src/app/components/import-data/import-data.component.ts
```diff
@@ -0,0 +1,168 @@
...
+
+        this.csvParseError = '';
+        let sourceColumns = Object.keys(result.data[0]);
+
+        let connection = this.connectionService.connect(this.dashboardState.getDatastoreType(),
+            this.dashboardState.getDatastoreHost());
+
+        connection.getTableNamesAndFieldNames(this.optionCollection.database.name,
+            ((response: any) => {
+                let destinatinColumns: string[] = response[this.optionCollection.table.name];
```
ThomasSchellenbergNextCentury: @aamduka The word "destination" in `destinatinColumns` is missing its `o`

src/app/components/import-data/import-data.component.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ImportDataComponent } from './import-data.component';
+
+import { DashboardService } from '../../services/dashboard.service';
+
+import { DashboardServiceMock } from '../../services/mock.dashboard-service';
+import { initializeTestBed } from '../../../testUtils/initializeTestBed';
+
+import { ImportDataModule } from './import-data.module';
+
+describe('Component: Import-Data', () => {
```
ThomasSchellenbergNextCentury, line 32: @aamduka I think we should add some tests for the `import` function to verify that it calls `connection.runImportQuery` with the expected arguments (like `importQuery`).

src/app/components/import-data/import-data.component.ts
```diff
@@ -0,0 +1,168 @@
...
+            this.csvParseError = "";
+        }
+
+        this.csvParseError = '';
+        let sourceColumns = Object.keys(result.data[0]);
+
+        let connection = this.connectionService.connect(this.dashboardState.getDatastoreType(),
+            this.dashboardState.getDatastoreHost());
+
+        connection.getTableNamesAndFieldNames(this.optionCollection.database.name,
```
ThomasSchellenbergNextCentury: @aamduka The `dataset` should already call `getTableNamesAndFieldNames` for you (when it's first created), so I don't think you need to call it here.  The `dataset` (and, transitively, the `optionCollection`) should have all the `datastores`, `databases`, `tables`, and `fields` you need already.

ThomasSchellenbergNextCentury:
@aamduka FYI You have some lint errors.  You can auto-fix some of them (and see the rest) by running `npm run-script lint`.

ThomasSchellenbergNextCentury:
@aamduka I'm concerned about the import behavior with very large files.  For example, I tested with some CSVs that are over 400MB (millions of rows).  They took many seconds to parse and I was concerned that my browser would run out of memory and crash.

At the very least, we should add a message and/or loading spinner that shows while the parser is running.

Additionally, we should consider one of the following:
1. Do nothing (keep everything the way it is).
2. Don't import very large files; show an error message instead.  This is not a good option.
3. Migrate the CSV parsing code to the Java server which can more easily handle large data processing tasks.  This would obviously be a lot of work and could introduce other unforeseen issues.
4. Chunk the CSV parsing.  Papaparse appears to have both `step` and `chunk` [config options](https://www.papaparse.com/docs#config) that we could use.  We'd still need to store all the imported rows in browser memory before sending them to the Server with `runImportQuery`, but I think that breaking up the parsing into smaller chunks would still help.  Plus this would be much easier to do than option 3 above.

Right now I'm leaning toward option 4.  Akmal and @chrislong what are your thoughts?

src/app/components/import-data/import-data.component.ts
```diff
@@ -0,0 +1,168 @@
...
+    public onImportClick() {
+        let file = this.inputFile.nativeElement.files[0];
+        this.papa.parse(file, { header: true,
+            complete: this.import.bind(this) });
+    }
+
+    import(result: any) {
+        if (result.errors.length > 0) {
+            this.csvParseError = `CSV parsing error at line ${result.errors[0].row}: ${result.errors[0].message}`;
+            return;
```
ThomasSchellenbergNextCentury: @aamduka If I get a parsing error on one of my file's rows (like "Too many fields" or "Too few fields"), I don't think the entire import should fail -- just that one row.  Maybe we should still show the error message but also continue with the import, and just remove all the rows in `result.data` that are invalid?

src/app/components/import-data/import-data.component.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    public onImportClick() {
+        let file = this.inputFile.nativeElement.files[0];
+        this.papa.parse(file, { header: true,
+            complete: this.import.bind(this) });
+    }
+
+    import(result: any) {
+        if (result.errors.length > 0) {
+            this.csvParseError = `CSV parsing error at line ${result.errors[0].row}: ${result.errors[0].message}`;
```
ThomasSchellenbergNextCentury: @aamduka Can we show the messages from ALL the `result.errors`, and not just the first one?  I get confused because I think my file only contains one invalid row when it actually contains multiple (since the other errors are hidden).

src/app/components/import-data/import-data.component.html
chrislong: That's a good point. I can't think of any other instructions.

chrislong:
I'm not sure what the best way to handle errors and big files is. It seems kinda wasteful to parse and error check on the client when the server will probably have to do that, too. But it would also be nice for the user to not have to wait until the entire file is uploaded to find out, for example, that there's an error on line 2. I like the idea of uploading in chunks, if that's feasible.

aamduka:
@ThomasSchellenbergNextCentury @chrislong All good points you guys raised. regarding big files and errors, after upload is complete, can we provide them a link to download an error report file which lists all the line numbers which failed to be imported. I also think that we should give them capability to remove all the table contents so that they would be able to re-run an upload. this might be a good feature in case where 5% of rows failed to import in a big file and user would like to re-run the upload after fixing the problematic rows.

I am taking the day off today but we can brainstorm on Monday if necessary

ThomasSchellenbergNextCentury:
@aamduka FYI I merged master into this branch and moved the changes from the Connection Service to the Component Library.

src/app/components/import-data/import-data.module.ts
aamduka: @ThomasSchellenbergNextCentury I don't have strong feeling one way or the other but I was thinking I want to give it a module scope so that I can access it from any component in the module. At this time, we only have one component so it doesn't really matter where I put it.

.gitignore
```diff
@@ -59,6 +59,9 @@ src/app/config/config.yaml
 src/app/config/config\.working\.json
 src/app/config/config\.working\.yaml
 
+# custom assets
+src/assets/custom/*
```
chrislong: What are custom assets and why wouldn't they be checked in?

e2e/app.po.ts
```diff
@@ -42,7 +42,7 @@ export class NeonGtdPage {
 
     async getVizPageInfo(element: ElementFinder): Promise<PageInfo | undefined> {
```
chrislong: This function could use some comments. I have no idea why it's munging the text the way it is.

quick-install/install.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+#neon installation on docker
+
+scriptdir="$( cd "$( dirname "$0")" && pwd )"
```
chrislong: It should be adequate to do `scriptdir="$(dirname $0)"`

src/app/components/contribution-dialog/contribution-dialog.component.html
```diff
@@ -8,7 +8,7 @@
     <div>The following organizations contributed to this display:</div>
     <div *ngFor="let item of data" class="contributor">
         <div class="contributor-logo">
-            <img src="./assets/images/contributor-logos/{{item.logo}}">
+            <img src="./assets/custom/{{item.logo}}">
```
chrislong: Would "contributors" be a more descriptive name for this than "custom"?

src/app/models/widget-option-collection.ts
```diff
@@ -227,7 +254,33 @@ export class OptionCollection {
...
+        if (this.datastores.length) {
+            let configuredDatastore;
+
+            if (this.tableKey) {
+                // The table key is either a tablekey string (datastore.database.table) or a unique ID in the dataset's tableKeyCollection.
+                const datasetTableKey = dataset.tableKeyCollection[this.tableKey];
+                configuredDatastore = DatasetUtil.deconstructTableOrFieldKeySafely(datasetTableKey || this.tableKey).datastore;
+            }
+
+            if (configuredDatastore) {
```
chrislong: This `if` could go inside the previous `if`, right? (And the scope of `configuredDatastore` reduced) That's the only way it'll be truthy, right?

src/app/components/document-viewer-single-item/document-viewer-single-item.component.ts
```diff
@@ -15,7 +15,7 @@
 import { Component, Inject, OnInit, OnDestroy } from '@angular/core';
 import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material';
 
-import { CoreUtil } from '../../library/core/core.util';
+import { CoreUtil } from 'component-library/dist/core/core.util';
```
chrislong: @ThomasSchellenbergNextCentury Why do all the imports for the component library start with "component-library/dist"? Why not just "component-library"?

chrislong (changes requested):
A few minor questions & suggestions

.gitignore
ThomasSchellenbergNextCentury: @chrislong This change was due to moving all the program-specific images out of the dashboard repo.  All the images were moved into the neon-assets repo.  You need to copy them from neon-assets into src/assets/custom/ and rebuild the dashboard.  I'm not sure why the changes are showing up here, though, since they should already be in master...

src/app/components/contribution-dialog/contribution-dialog.component.html
ThomasSchellenbergNextCentury: @chrislong This change was due to moving all the program-specific images out of the dashboard repo.  The assets in the "custom" folder can be anything, not just contributors.

src/app/components/document-viewer-single-item/document-viewer-single-item.component.ts
ThomasSchellenbergNextCentury: @chrislong The imports from the component-library are a little wonky right now.  Eventually, we'll want the files from the "core" folder and each individual visualization to be separate NPM modules.  Since I knew we were going to change the imports in the near future, I just did it "the way that worked" for now.

aamduka:
@ThomasSchellenbergNextCentury @chrislong I have incorporated the suggestions from the code review discussions. please review when you get a chance and let me know if we need to do further improvements

chrislong (commented):
No comments on the code. FYI, I won't have a chance to test it until Thurs, probably.

chrislong:
LGTM

ThomasSchellenbergNextCentury (approved):
This looks great!  Nice job, Akmal!

PR #374, opened 2020-04-19
Fixed pie chart visualization.

ThomasSchellenbergNextCentury:
A grouped pie chart will now show a separate slice for each X + group combination, colored by the group.  Clicking on a slice will filter by the X value.


ScottWarner:
This does fix the coloring issue.
There was a minor compile error, but I added a commit to fix that

ScottWarner (approved):

ThomasSchellenbergNextCentury:
Awesome!  Thanks!

